PR #1, opened 2026-03-02
checkpoint Draft PR

luke-venk:


backend_rust/src/camera.rs
```diff
@@ -0,0 +1,7 @@
+// We currently are only using 2 ground cameras.
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum CameraId {
+    Camera1,
```
owens3364: Can name these FieldLeft and FieldRight or something indicating their position

PR #2, opened 2026-03-04
feat: initialization of backend pipeline, some reorganization

luke-venk:
I added the following related to the pipeline:
- queue: wrapper around `crossbeam_channel` to serve as a thread-safe queue the producer and consumer can read from
- producer: does nothing right now, but will handle ingesting frames from cameras and enqueuing
- consumer: does nothing right now, but will dequeue frames and run CV on them
- pipeline: handles creating the queue and threads for the producers/consumers to run on

I also added the super bare bones for an Axum server. Right now it just performs a health check, no need to review yet but pushed just so you can see it.

I added unit tests for the queue and server as well. I added dummy code for the producers to just make frames with placeholder data and the consumers to print out the frame's placeholder data. Running `cargo run` will verify that two pipelines are made and the program will print stuff out. _drop earliest behavior_ can be verified by making the producer run significantly faster than the consumers.

### DIFF
--- a/backend/src/schemas/frame.rs
+++ b/backend/src/schemas/frame.rs
@@ -1,23 +1,23 @@
 /**
  * Each camera frame will consist of bytes representing the actual
- * image as well as metadata (Context).
+ * image as well as timestamp (Context).
  */
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct Context {
     // TODO: replace placeholder
-    metadata: u64,
+    timestamp: u64,
 }
 
 impl Context {
-    pub fn new(metadata: u64) -> Self {
+    pub fn new(timestamp: u64) -> Self {
         Self {
-            metadata
+            timestamp
         }
     }
 
-    pub fn metadata(&self) -> u64 {
-        self.metadata
+    pub fn timestamp(&self) -> u64 {
+        self.timestamp
     }
 }
 
@@ -54,6 +54,6 @@ mod tests {
         let data = vec![1, 2, 3, 4];
         let frame = Frame::new(data, Context::new(34151));
         
-        assert_eq!(frame.context().metadata(), 34151);
+        assert_eq!(frame.context().timestamp(), 34151);
     }
 }
--- a/backend/src/computer_vision/contour.rs
+++ b/backend/src/computer_vision/contour.rs
@@ -4,6 +4,6 @@ pub fn contour(frame: Frame) -> Frame {
     // TODO: implement actual logic
     
     let data = frame.data();
-    let new_metadata = frame.context().metadata() + 1;
-    Frame::new(data.to_vec(), Context::new(new_metadata))
+    let new_timestamp = frame.context().timestamp() + 1;
+    Frame::new(data.to_vec(), Context::new(new_timestamp))
 }
--- a/backend/src/computer_vision/forward_downsampled_copy.rs
+++ b/backend/src/computer_vision/forward_downsampled_copy.rs
@@ -4,6 +4,6 @@ pub fn forward_downsampled_copy(frame: Frame) -> Frame {
     // TODO: implement actual logic
     
     let data = frame.data();
-    let new_metadata = frame.context().metadata() + 1;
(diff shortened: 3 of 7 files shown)


backend/src/hardware/frame.rs
```diff
@@ -0,0 +1,35 @@
+/**
+ * Each camera frame will consist of metadata and the actual image.
+ */
+
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Frame {
+    // TODO: add image data
```
owens3364: Just add a Vec<u8> - that's how we'll receive it

backend/src/server/server.rs
```diff
@@ -0,0 +1,109 @@
...
+        "javelin" => EventType::Javelin,
+        _ => return Err(ApiError::InvalidEvent),
+    };
+    Ok(Json(SpecifyEventTypeResponse { event_type }))
+}
+
+pub fn create_app() -> Router {
+    Router::new()
+        .route("/health", get(health_check))
+        .route("/event_type", post(specify_event_type))
```
owens3364: Add a `get` for the same `/event_type` route that lets you query what event type the server is currently set to

backend/src/main.rs
```diff
@@ -0,0 +1,35 @@
...
+    // Build the Axum router.
+    let app = create_app();
+
+    // Build the shutdown signal: TODO
+    
+    // Start the Axum server.
+    connect_to_server(app, "0.0.0.0:3000").await;
+    println!("Server running on http://localhost:3000");
+
+    // Once the server is stopped, also stop the pipelines gracefully and block
```
owens3364: Again, not sure how this code would be reached, I think I'm missing something?

owens3364 (changes requested):

backend/src/hardware/frame.rs
luke-venk: I'll add a Vec, that makes sense.

Right now my idea was that each camera would have their own pipeline with their own frames. Upon further consideration I think it makes sense that Frame has both of the camera's images, but wouldn't that make ingestion of the left camera tightly coupled to the ingestion of the right camera?

backend/src/server/server.rs
```diff
@@ -0,0 +1,109 @@
...
+    event_type: String,
+}
+
+#[derive(Serialize)]
+struct SpecifyEventTypeResponse {
+    event_type: EventType,
+}
+
+// Ensure handler knows how to return custom ApiError as part of HTTP response.
+impl IntoResponse for ApiError {
```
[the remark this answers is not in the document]
owens3364: IMO - see above, remove

backend/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+edition = "2024"
+
+[dependencies]
+axum = "0.8.8"
+crossbeam-channel = "0.5.15"
+serde = { version = "1.0.228", features = ["derive"]}
+serde_json = "1.0"
+tokio = { version = "1.49.0", features = ["full"]}
+
+[dev-dependencies]
```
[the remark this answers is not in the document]
owens3364: Ahh ok

owens3364 (changes requested):
This is pretty dope. I’ll review a bit more thoroughly later but I like a lot of what I see.

Couple notes

1) print statements in pipeline stages (or any multithreaded context) are gonna slow things down and get mangled. I understand the motivation, if two things print to the console at the same time it’ll just look like shit.
2) right now this runs on 3000 - let’s make it run on 5000. The next JS server runs on 3000 so if we want to run both at same time and integration test we can’t do that right now

backend/src/schemas/frame.rs
```diff
@@ -0,0 +1,59 @@
+/**
+ * Each camera frame will consist of bytes representing the actual
+ * image as well as metadata (Context).
+ */
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Context {
+    // TODO: replace placeholder
```
owens3364: Rename metadata to timestamp

Issue #3, opened 2026-03-04
Implement Camera Ingest


luke-venk:
## Description
Figure out how to get footage from the cameras into our pipeline. We should try to leverage our existing recording binaries. Load data into the `Frame` struct which is sent into the pipeline for processing.

Right now there are 2 types of backend builds, `fake` and `real`. See README.bazel.md for more info, but this change only needs to be implemented for `real` mode when we actually spin up the pipelines. This will also require testing with the actual cameras.

## Relevant files
- `backend/src/bin/record_from_both_cameras.rs`
- `backend/src/camera_ingest/camera_ingest.rs`
- `backend/src/schemas/frame.rs`

RushilRandhar:
This is largely covered by PR #78 (camera ingest replay pipeline + manifest + assignment resolver). That PR brings in the ingest infrastructure originally developed on \`test-camera-ingest\` and lands it on main, including a working \`ingest_frames\` (replacing the \`todo!()\` currently on main), the recording → replay flow, and the centralized left/right assignment resolver.

Suggest closing this once #78 lands. Flagging early so it's on the radar.

luke-venk:
Closed by #78

PR #10, opened 2026-03-09
Camera ingest 

grao23:
Added camera ingest by assigning each ingest thread a left or right camera slot, defined camera_ingest_config to hold per-camera settings. 

Added rough workflow of how I was planning to build out the actual frame ingest using Aravis. 

Lmk what y'all's thoughts are. It is running, but I would like to know where I can improve with this/ if the workflow thought is correct, so I can start building that out.

### DIFF
--- a/backend/src/camera_ingest/camera_ingest.rs
+++ b/backend/src/camera_ingest/camera_ingest.rs
@@ -1,12 +1,126 @@
-use crate::schemas::{Frame, Context};
+use std::thread;
+use std::time::{Duration, Instant};
+
+use aravis::{BufferStatus, CameraExt, StreamExt};
+use crossbeam::channel::Sender;
+
+use crate::camera_ingest::camera_ingest_helpers::{
+    buffer_to_frame, configure_camera, create_stream_and_queue_buffers, initialize_aravis,
+    open_camera,
+};
+use crate::schemas::camera_ingest_config::CameraIngestConfig;
+use crate::schemas::{Context, Frame};
 
 // Ingests frames from the cameras using the GigEVision API, and enqueues
 // the frames into the camera ingest sender for the camera's pipeline to
 // begin processing.
-pub fn ingest_frames(tx: crossbeam::channel::Sender<Frame>){
-    // TODO(#3): Implement Camera Ingest.
-    
-    let data = vec![1, 2, 3, 4];
-    let context = Context::new(1);
-    let _ = tx.send(Frame::new(data, context));
+pub fn ingest_frames(tx: Sender<Frame>, config: CameraIngestConfig) {
+    let config = config.validate().expect("invalid camera ingest config");
+    let start_time = Instant::now();
+    let mut frames_sent = 0usize;
+
+    if config.use_fake_interface {
+        loop {
+            if let Some(limit) = config.max_frames {
+                if frames_sent >= limit {
+                    break;
+                }
+            }
+
+            if let Some(seconds) = config.max_duration_s {
+                if start_time.elapsed() >= Duration::from_secs_f64(seconds) {
+                    break;
+                }
+            }
+
+            let data = vec![1, 2, 3, 4];
+            let context = Context::new(1234);
+            if tx.send(Frame::new(data, context)).is_err() {
+                break;
+            }
+            frames_sent += 1;
+            thread::sleep(Duration::from_millis(10));
+        }
+
+        return;
+    }
+
+    initialize_aravis();
+    let camera = open_camera(&config);
+    configure_camera(&camera, &config);
+    let stream = create_stream_and_queue_buffers(&camera, config.num_buffers);
+
+    camera
+        .start_acquisition()
+        .expect("Failed to start acquisition.");
+
+    loop {
+        if let Some(limit) = config.max_frames {
+            if frames_sent >= limit {
+                break;
+            }
+        }
+
+        if let Some(seconds) = config.max_duration_s {
+            if start_time.elapsed() >= Duration::from_secs_f64(seconds) {
+                break;
+            }
+        }
+
+        let buffer = match stream.timeout_pop_buffer(config.timeout_ms * 1000) {
+            Some(buffer) => buffer,
+            None => continue,
+        };
+
+        match buffer.status() {
+            BufferStatus::Success => {
+                let frame = buffer_to_frame(&buffer);
+
+                if tx.send(frame).is_err() {
+                    stream.push_buffer(buffer);
+                    break;
+                }
+
+                frames_sent += 1;
+            }
+            status => {
+                eprintln!("Buffer status: {:?}", status);
+            }
+        }
+
+        stream.push_buffer(buffer);
+    }
+
+    let _ = camera.stop_acquisition();
+}
+
+#[cfg(test)]
+mod tests {
+    use crossbeam::channel::bounded;
+
+    use super::ingest_frames;
+    use crate::schemas::camera_ingest_config::{CameraIngestConfig, Resolution};
+
+    #[test]
+    fn fake_ingest_stops_after_max_frames() {
+        let (tx, rx) = bounded(4);
+        let config = CameraIngestConfig {
+            device_id: String::new(),
+            exposure_time_us: 1000.0,
+            frame_rate_hz: 30.0,
+            resolution: Resolution::UHD4K,
+            aperture: None,
+            enable_ptp: false,
+            use_fake_interface: true,
+            num_buffers: 2,
+            timeout_ms: 100,
+            max_frames: Some(2),
+            max_duration_s: None,
+        };
+
+        ingest_frames(tx, config);
+
+        let frames: Vec<_> = rx.try_iter().collect();
+        assert_eq!(frames.len(), 2);
+    }
 }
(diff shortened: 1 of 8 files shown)


luke-venk (changes requested):

luke-venk (changes requested):

owens3364 (changes requested):
Would like to hear your thoughts on this @luke-venk but I think
1. We should specify camera properties via CLI args
2. We should discover camera devices on the network and offer them to the user via CLI, not ask for hardcoded IDs via environment variable

Basically I envision a couple of utility binaries for gathering data

```bash
$ ./discover_cameras
[Camera 1 ID] [IP addr] [MAC] [Hardware type] [Any other relevant discovery details]
[Camera 2 ID] [IP addr] [MAC] [Hardware type] [Any other relevant discovery details]
$ ./record_from_cameras --camera camera1id --camera camera2id --resolution 4k --aperature f/8 --shutter_speed 25.4 --save_recordings_dir=~/Downloads/mydir
Recording...
Recording...
Recording...
<Ctrl + C>
Saved to ~/Downloads/mydir/recordingcam1.mp4 ~/Downloads/mydir/recordingcam2mp4
$ 
```
- PTP can be enabled automatically if multiple cameras are selected for recording
- Exposure time is restricted to the range of 25.4µs to 10s according to camera specs and we should validate that and support the full range
- We should look into if aperture and F stop can be set by aravis or not (probably at least one of them can?)

This PR can be utility functions to serve the above goal, and we can have several rust_binary targets in Bazel with their own main.rs files that create discover_cameras and record_from_cameras and such.

But I do think we should move away from env vars and toward a CLI-native approach.

backend/src/config/field_left.rs
```diff
@@ -0,0 +1,8 @@
+ARAVIS_LEFT_CAMERA_ID=
```
luke-venk: I don't think that Rust (.rs) files can set env vars like this. For one, this is not valid Rust code.

Pretty sure the convention for setting env vars is you have a .env file, and then in your code at startup you need to load the env vars (you can use a crate like [dotenvy](https://crates.io/crates/dotenvy)), and then you can use them.

Not high priority since we seemed to have moved away from env vars, but you can delete these files for now.

backend/src/camera_ingest/camera_ingest.rs
```diff
@@ -1,12 +1,81 @@
...
 use crate::schemas::{Frame, Context};
+use aravis::prelude::*;
+use aravis::{AcquisitionMode, Aravis, Buffer, BufferStatus, Camera, ExposureMode};
+use crossbeam::channel::Sender; 
+use crate::camera_ingest::camera_ingest_helpers::{
+    buffer_to_frame, configure_camera, create_stream_and_queue_buffers, initialize_aravis,
+    open_camera,};
+use crate::camera_ingest_config::CameraIngestConfig;
+
+//Intializes Aravis once
```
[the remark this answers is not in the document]
luke-venk: If `Once` is truly not needed, please delete the code instead of commenting it out. Also, please do the same for all other instances of `Once` in other files and resolve the associated comments.

backend/src/camera_ingest/camera_ingest.rs
```diff
@@ -1,12 +1,79 @@
...
+    //Define stream.
+    let stream = create_stream_and_queue_buffers(&camera, config.num_buffers); 
+
+    //
+    camera
+        .start_acquisition()
+        .expect("Failed to start it");
+
+    //Pulling buffers from camera stream
+loop {
```
luke-venk: nit: fix indentation

backend/src/camera_ingest/camera_ingest.rs
```diff
@@ -1,12 +1,79 @@
...
+    }
+
+    initialize_aravis();
+    //Define camera. 
+    let camera = open_camera(&config); 
+    configure_camera(&camera, &config);
+    //Define stream.
+    let stream = create_stream_and_queue_buffers(&camera, config.num_buffers); 
+
+    //
```
luke-venk: nit: please finish comment

luke-venk (changes requested):

backend/src/cli.rs
```diff
@@ -0,0 +1,47 @@
+use clap::{Args, Parser, Subcommand};
```
luke-venk: This file should probably go inside `tools/` alongside the `env.rs` file, right?

backend/src/bin/discover_cameras.rs
```diff
@@ -0,0 +1,24 @@
...
+            "id={} ip={} physical_id={} vendor={} model={} protocol={}",
+            cam.id,
+            cam.address,
+            cam.physical_id,
+            cam.vendor,
+            cam.model,
+            cam.protocol,
+        );
+    }
+}
```
luke-venk: nit: add newline at end of file

I also made this mistake on a previous PR: https://github.com/luke-venk/mjolnir/pull/13#discussion_r2976708542

backend/src/bin/record_from_cameras.rs
```diff
@@ -0,0 +1,35 @@
...
+                .expect("invalid camera ingest config");
+
+                ingest_frames(tx, config);
+            }
+        }
+        Commands::DiscoverCameras => {
+            eprintln!("use //backend:discover_cameras for camera discovery");
+        }
+    }
+}
```
luke-venk: nit: newline

backend/src/camera_ingest/mod.rs
```diff
@@ -1,3 +1,4 @@
 pub mod camera_ingest;
-
+pub mod camera_ingest_helpers;
+pub mod discovery;
 pub use camera_ingest::ingest_frames;
```
luke-venk: nit: newline

backend/src/tools/env.rs
```diff
@@ -0,0 +1,148 @@
...
+        unsafe {
+            env::set_var(key, "banana");
+        }
+        assert!(read_env_bool(key, true));
+        assert!(!read_env_bool(key, false));
+        unsafe {
+            env::remove_var(key);
+        }
+    }
+}
```
luke-venk: nit: newline

backend/src/tools/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod env;
```
luke-venk: nit: newline

backend/src/bin/record_from_cameras.rs
```diff
@@ -0,0 +1,35 @@
...
+                    num_buffers: args.num_buffers,
+                    timeout_ms: args.timeout_ms,
+                }
+                .validate()
+                .expect("invalid camera ingest config");
+
+                ingest_frames(tx, config);
+            }
+        }
+        Commands::DiscoverCameras => {
```
luke-venk: I might be misunderstanding, but why would the Command ever be `DiscoverCameras` if this binary is called? If we wanted to run `discover_cameras`, wouldn't it just call the other binary? Like why wouldn't this match statement also be needed in `discover_cameras.rs`?

backend/src/bin/record_from_cameras.rs
```diff
@@ -0,0 +1,35 @@
...
+use mjolnir::cli::{Cli, Commands};
+use mjolnir::schemas::camera_ingest_config::CameraIngestConfig;
+
+fn main() {
+    let cli = Cli::parse();
+
+    match cli.command {
+        Commands::RecordFromCameras(args) => {
+            for camera in args.cameras {
+                let (tx, _rx) = unbounded();
```
luke-venk: I think that if `_rx` isn't used, just call it `_`

backend/src/bin/record_from_cameras.rs
```diff
@@ -0,0 +1,35 @@
...
+                    aperture: args.aperture,
+                    enable_ptp: false,
+                    use_fake_interface: args.use_fake_interface,
+                    num_buffers: args.num_buffers,
+                    timeout_ms: args.timeout_ms,
+                }
+                .validate()
+                .expect("invalid camera ingest config");
+
+                ingest_frames(tx, config);
```
luke-venk: if `ingest_frames()` is now called in this binary, how does that impact line 37 in `pipeline.rs` where the same function is also called? Do we need to call it there anymore? Do we need to update its functionality?

luke-venk (changes requested):

backend/src/bin/record_from_cameras.rs
owens3364: I haven't actually looked at this in detail but it is concerning to me that we are just throwing away the _rx object. They means that once something is consumed by tx it just disappears into the void and no one else can receive it. At that point, there's no point in having a crossbeam channel. What's going on here?

luke-venk: Oh yeah that’s a really good point. I think this compiles rn but won’t actually work. 

@grao23 Previously, all the tx and rx were initialized in pipeline.rs and then passed to camera_ingest.rs. With this refactoring into different binaries, that code will have to also be updated. 

Kinda related comment: https://github.com/luke-venk/mjolnir/pull/10#discussion_r2997651161

grao23: The record_from_cam is throwing some errors so when I had pushed it, I had left it as a placeholder till I finish reconnecting where the frames are supposed to go, but yes yall are right this is not how it should be. Working on fixing this now.

luke-venk:
Closing this because majority of these changes were covered in [PR 25](https://github.com/luke-venk/mjolnir/pull/25). Remaining changes will be covered in #3.

PR #15, opened 2026-03-24
Axum Serves Embedded Frontend Static Exports

luke-venk:
Following updates:
* I split backend into a rust library, and 2 executables (dev and deploy), to clean up how the servers run
* deploy.rs is the entrypoint for the final product, and it uses [rust-embed](https://docs.rs/crate/rust-embed/latest) to embed the frontend static exports into the binary, and uses [axum-embed](https://docs.rs/axum-embed/0.1.0/axum_embed/) to serve the files
* dev.rs just handles CORS and basic serving the API
* I added a deploy target that ensures the frontend static exports and backend deploy targets are built, and then executes the binary

You can verify this works by building
`bazel run --config=release //deploy:mjolnir`  
and opening `localhost:5001` in your browser.

You can really prove that there is no reliance on a frontend out/ directory, and that assets are truly embedded by moving the final product (`bazel-bin/backend/deploy`) to any directory on your PC and just executing it.


luke-venk:
It should be noted that the placeholder pictures no longer load. Since these images are from online and not in the static export, they don't load. I'm not really sure if this will be an issue or not for when we display the actual pictures from the camera.

If this indeed is an issue, this can be a spinoff ticket. lmk what you think please

image

MODULE.bazel.lock
owens3364, line 1: I thought we agreed that this file shouldn't be committed? Or am I missing something?

backend/BUILD.bazel
```diff
@@ -1,14 +1,20 @@
-load("@rules_rust//rust:defs.bzl", "rust_binary", "rust_test")
+load("@rules_rust//rust:defs.bzl", "rust_binary", "rust_library", "rust_test")
 
-rust_binary(
-    name = "mjolnir",
-    srcs = glob(["src/**/*.rs"]),
+# Define a library that compiles the source code and can be used by
+# both the dev and deploy binaries. Can be optimized using command
+# line arguments.
+rust_library(
```
owens3364, line 6: The rustc_flags should probably be a part of the binary, not the library, right? and then you could still strip symbols?

owens3364 (changes requested):
I'm a bit confused by the verbiage on the deploy things. It looks like 'deploy' actually means, 'run a production build', yes?

I would do the following
- drop the bash scripts and bazel bash tooling (pretty sure you don't need them)
- rename 'deploy' things to be 'prod' things
- take 'dev.rs' and 'prod.rs' out of the bin folder and put them in your src/ folder (you could do this for server.rs too if you wanted but ¯\_(ツ)_/¯)
- Rename 'dev.rs' and 'prod.rs' to 'main.dev.rs' and 'main.prod.rs' and just select different ones for different bazel binaries by glob `exclude`ing the entrypoint files that you don't want

I would start there and then running dev vs prod should be cleaner and more obvious as to what is what and require less tooling

deployment usually means 'generate a set of files and put them somewhere accessible to others' which is not quite what this does.

luke-venk:
> I'm a bit confused by the verbiage on the deploy things. It looks like 'deploy' actually means, 'run a production build', yes?
> 
> I would do the following
> 
> * drop the bash scripts and bazel bash tooling (pretty sure you don't need them)
> * rename 'deploy' things to be 'prod' things
> * take 'dev.rs' and 'prod.rs' out of the bin folder and put them in your src/ folder (you could do this for server.rs too if you wanted but ¯_(ツ)_/¯)
> * Rename 'dev.rs' and 'prod.rs' to 'main.dev.rs' and 'main.prod.rs' and just select different ones for different bazel binaries by glob `exclude`ing the entrypoint files that you don't want
> 
> I would start there and then running dev vs prod should be cleaner and more obvious as to what is what and require less tooling
> 
> deployment usually means 'generate a set of files and put them somewhere accessible to others' which is not quite what this does.

You're totally right about the "deploy" vs. "prod" thing, I was mixing up terminology. Also, yeah the bash script is totally useless. I originally did this because I thought this was the nicest way to just automate everything but it didn't actually add value and it was confusing. Undoing now.

The `main.dev.rs` and `main.prod.rs` points also make sense. I originally thought that all different binaries should go in `src/bin` but [Rust convention](https://doc.rust-lang.org/rust-by-example/cargo/conventions.html) seems to be putting default binaries in `src/` directly (since dev and prod are variants of the same program) and additional, unrelated binaries in `src/bin` (like the ones Gautam will add for discovery and recording). Not that convention technically matters since we're not using Cargo but still good practice.

I'm don't really understand what you're saying about `server.rs`. Are you saying it shouldn't be in `backend/src/server`? It's not a standalone executable, I just put server related code there.

MODULE.bazel.lock
luke-venk: This is the comment we previously discussed this on: https://github.com/luke-venk/mjolnir/pull/13/changes#r2972670652

The official [Bazel Lockfile](https://bazel.build/external/lockfile#best-practices) docs says best practice is to commit it to ensure consistent development environments across teammates.

I know in the trel3 repo it's not there, but isn't this the same idea as committing something like `pnpm-lock.yaml`, which is checked in in trel3?

I can remove the file if you'd like. I'm not sure what the consequences are, although it's probably fine since it's not in the trel3 repo. I thought it was good practice though.

backend/BUILD.bazel
luke-venk: I thought the same thing at first. I put the flags on the rust_binary for prod and not the rust_library originally.

See the [comment from a previous PR](https://github.com/luke-venk/mjolnir/pull/13#discussion_r2972686931) in which I compared the sizes of the binaries with the flags and without. Note that the optimized size was 1.1 MB and the normal was 5.2 MB.

But when I didn't add the rustc_flags on the library, and only on the binary, it would compile the library without the optimizations, which is like 95% of the code in the binary, although it compiled the binary with optimizations. This made the binary size like 4 MB instead of the expected 1.1 MB. I added the conditional flags back to the library and it fixed it.

luke-venk:
> It should be noted that the placeholder pictures no longer load. Since these images are from online and not in the static export, they don't load. I'm not really sure if this will be an issue or not for when we display the actual pictures from the camera.
> 
> If this indeed is an issue, this can be a spinoff ticket. lmk what you think please
> 
> image

This was a one line fix. I used ChatGPT to figure this out but basically `next/image` in `frontend/app/components/image-gallery.tsx` usually expects Next.js server-side image handling, but embedded that into the binary no longer has this. So changing `next.config.ts` to use `unoptimized` because Next no longer tries using its image optimization pipeline and just renders normal URLs.

Relevant docs here: https://nextjs.org/docs/messages/export-image-api

Your instinct was totally right, nice. Now, if the build is static exported, it will use regular image URLs instead of Next's image optimization pipeline.

MODULE.bazel.lock
owens3364: Ah, yes, bazel does say it's best practice to commit it, but after trying it in trel3 it sucked.
The argument for committing it is that it ensures that everyone's bazel repository cache resolves to the exact same things. Or, it _should_...

What actually happens is that different platforms will resolve different toolchains for their usages, which means that different users on different platforms will include lockfile diffs in their PRs even if no bazel files are changed in that PR.

Bazel is trying to get to a place where every toolchain repository must be reproducible so that this issue doesn't happen, but they aren't there yet.

[This buried bit of documentation](https://bazel.build/rules/lib/builtins/repository_ctx) describes the `reproducible` parameter of repository metadata that actually indicates if a repository will produce the same MODULE.bazel.lock results for the same exact inputs (I *believe* in a platform-agnostic way but the docs are missing that detail I think).

But yeah, we can't guarantee that our dependencies and toolchain repositories will make the same MODULE.bazel.lock even when everything else is held equal.

To contrast this with `pnpm-lock.yaml`, npm dependencies are js/ts, and js/ts are platform-agnostic languages that run inside vms that are remarkably consistent environments. Javascript runs the same way regardless of os, cpu arch, etc. This isn't aaaaassss true when you account for interaction with things outside the [V8 engine](https://en.wikipedia.org/wiki/V8_(JavaScript_engine)) or issues with browsers not respecting standards (ahem internet explorer) but it's generally accurate. There's never a different pnpm package for windows vs mac browsers.

Bazel toolchains are inherently different on different environments though. On linux we use the mold linker, but on mac you literally can't run that, and the best option is to use the linker bundled with Apple clang. So the required behavior is resolving to different tools per-platform, whereas pnpm doesn't give a shit about your platform.

backend/BUILD.bazel
owens3364: Damn, that's super interesting. We may need to make adjustments to our build processes in trel3.
Can you add back strip symbols to rust_library, but make it `-C strip=debuginfo`? Linking with the binary should still succeed and it should improve the binary size a bit (maybe not by much).

owens3364:
> > I'm a bit confused by the verbiage on the deploy things. It looks like 'deploy' actually means, 'run a production build', yes?
> > I would do the following
> > 
> > * drop the bash scripts and bazel bash tooling (pretty sure you don't need them)
> > * rename 'deploy' things to be 'prod' things
> > * take 'dev.rs' and 'prod.rs' out of the bin folder and put them in your src/ folder (you could do this for server.rs too if you wanted but ¯_(ツ)_/¯)
> > * Rename 'dev.rs' and 'prod.rs' to 'main.dev.rs' and 'main.prod.rs' and just select different ones for different bazel binaries by glob `exclude`ing the entrypoint files that you don't want
> > 
> > I would start there and then running dev vs prod should be cleaner and more obvious as to what is what and require less tooling
> > deployment usually means 'generate a set of files and put them somewhere accessible to others' which is not quite what this does.
> 
> You're totally right about the "deploy" vs. "prod" thing, I was mixing up terminology. Also, yeah the bash script is totally useless. I originally did this because I thought this was the nicest way to just automate everything but it didn't actually add value and it was confusing. Undoing now.
> 
> The `main.dev.rs` and `main.prod.rs` points also make sense. I originally thought that all different binaries should go in `src/bin` but [Rust convention](https://doc.rust-lang.org/rust-by-example/cargo/conventions.html) seems to be putting default binaries in `src/` directly (since dev and prod are variants of the same program) and additional, unrelated binaries in `src/bin` (like the ones Gautam will add for discovery and recording). Not that convention technically matters since we're not using Cargo but still good practice.
> 
> I'm don't really understand what you're saying about `server.rs`. Are you saying it shouldn't be in `backend/src/server`? It's not a standalone executable, I just put server related code there.

Ok sweet glad all that is good.

Re server.rs, I just meant that you could move it from src/server/server.rs to src/server.rs if you wanted to, not that you necessarily needed to :)

backend/src/main.dev.rs
```diff
@@ -0,0 +1,37 @@
...
+use tower_http::cors::{Any, CorsLayer};
+
+// In dev mode, the backend can serve the API via command line, and it will
+// also serve the Next.js server, so it will need CORS. It will not have any
+// embedded assets.
+pub fn create_dev_app() -> Router {
+    // Set up CORS layer to allow cross-origin sharing for integration mode.
+    // Next.js requests will come from port 3000.
+    let cors = CorsLayer::new()
+        .allow_origin(HeaderValue::from_static("http://localhost:3000"))
```
owens3364: This isn't a huge deal but I believe this will mean that, if I demo on my laptop in dev mode, if someone else on my network wants to see the app, it will not work on their machine, because while the origin on my laptop is 'localhost:3000', the origin on someone else's laptop would be '255.255.255.255:3000' where the 255s are my IP address.
In dev mode it's probably fine to just allow any origin. I believe you can literally do `allow_origin(Any)`

backend/src/main.dev.rs
```diff
@@ -0,0 +1,37 @@
...
+        .allow_methods([Method::GET, Method::POST])
```
owens3364: Nit: Add HEAD as well, sometimes browsers use it before making a full GET request

backend/BUILD.bazel
```diff
@@ -20,14 +26,67 @@ rust_binary(
         "@crates//:tokio",
         "@crates//:tower-http",
     ],
+    visibility = ["//visibility:public"],
```
owens3364: Does it need to be public? I don't believe it's referenced by anything outside this file, right?

backend/BUILD.bazel
```diff
@@ -20,14 +26,67 @@ rust_binary(
...
 )
 
+# Note, the main reason for splitting into dev and prod
+# binaries is to handle differences in how the applications serve
+# the frontend. Optimization can technically occur for either target
+# using `--config=release` (see .bazelrc).
+
+# Dev mode for the backend doesn't have to be optimized. In this mode,
+# the backend can serve the API via command line, and it will also serve
```
owens3364: It doesn't 'serve' the Next.js server, it runs alongside it. But yes, due to the API calls being at a different origin than the origin serving the web assets, it will need CORS

backend/BUILD.bazel
```diff
@@ -20,14 +26,67 @@ rust_binary(
...
+        "EMBEDDED_FRONTEND_DIR": "$(execpath //frontend:static_exports)",
+    },
+    rustc_flags = select({
+        "//:release": [
+            "-Cpanic=abort",
+            "-Cstrip=symbols",
+        ],
+        "//conditions:default": ["-Cstrip=symbols"],
+    }),
+    visibility = ["//visibility:public"],
```
owens3364: Don't think this needs to be public

backend/BUILD.bazel
```diff
@@ -20,14 +26,67 @@ rust_binary(
...
+        "@crates//:rust-embed",
+    ],
+    compile_data = ["//frontend:static_exports"],
+    # At compile-time, this env var tells our source code exactly where
+    # Bazel placed the frontend outputs, so rust-embed can embed it
+    # into the binary directly.
+    rustc_env = {
+        "EMBEDDED_FRONTEND_DIR": "$(execpath //frontend:static_exports)",
+    },
+    rustc_flags = select({
```
owens3364: You can just do
```starlark
["-Cstrip=symbols}] + select({
  "//:release": [
    "-Cpanic=abort",
    "-Ccodegen-units=1",
    "-Copt-level=3",
  ],
  "//conditions:default": [],
}),
```
regardless of style, I think the codegen units and optimization level settings belong here, yes?

backend/BUILD.bazel
```diff
@@ -20,14 +26,67 @@ rust_binary(
...
+    visibility = ["//visibility:public"],
+)
+
+# Unit tests to test the backend library.
 rust_test(
     name = "tests",
-    crate = ":mjolnir",
+    crate = ":backend_lib",
     deps = [
         "@crates//:axum",
```
owens3364: Axum is already a dep of backend_lib, I don't believe you need to mark it as a dep here too.

frontend/app/layout.tsx
```diff
@@ -2,8 +2,8 @@ import type { Metadata } from "next";
 import "./globals.css";
 
 export const metadata: Metadata = {
```
owens3364, line 4: Damn you think your browser metadata is better than mine huh?

It can stay...... love me some b2b saas revolution...

frontend/BUILD.bazel
```diff
@@ -121,5 +121,6 @@ js_run_binary(
     env = {
         "NEXT_PUBLIC_API_BASE_URL": "",
         "NEXT_PUBLIC_STATIC_EXPORT": "true",
-    }
-)
\ No newline at end of file
+    },
+    visibility = ["//visibility:public"],
```
owens3364, line 125: Does this need to be public? I don't believe anything references it?

owens3364 (changes requested):

MODULE.bazel.lock
luke-venk: Wow dude that's really interesting. Thanks for explaining that for me. I'll remove it.

backend/BUILD.bazel
luke-venk: I wanted to recreate what I previously found for concrete evidence, and what I did again didn't have as strong a conclusion as before.

## 1. Building release version with flags on binary but not library (2.5 MB)
```
➜  mjolnir git:(axum-serves-static-files) ✗ bazel build --config=release //backend:prod
INFO: Analyzed target //backend:prod (677 packages loaded, 14770 targets configured).
INFO: Found 1 target...
Target //backend:prod up-to-date:
  bazel-bin/backend/prod
INFO: Elapsed time: 47.006s, Critical Path: 43.19s
INFO: 2088 processes: 1286 internal, 799 darwin-sandbox, 3 local.
INFO: Build completed successfully, 2088 total actions
➜  mjolnir git:(axum-serves-static-files) ✗ ls -lh bazel-bin/backend/prod
-r-xr-xr-x@ 1 Luke  staff   2.5M Mar 27 08:20 bazel-bin/backend/prod
```

## 2. Building unoptimized version (4.9 MB)
```
➜  mjolnir git:(axum-serves-static-files) ✗ bazel build //backend:prod
WARNING: Build options --@@rules_rust+//rust/settings:lto and --compilation_mode have changed, discarding analysis cache (this can be expensive, see https://bazel.build/advanced/performance/iteration-speed).
INFO: Analyzed target //backend:prod (0 packages loaded, 3253 targets configured).
INFO: Found 1 target...
Target //backend:prod up-to-date:
  bazel-bin/backend/prod
INFO: Elapsed time: 14.062s, Critical Path: 12.53s
INFO: 1606 processes: 493 action cache hit, 1032 internal, 572 darwin-sandbox, 2 local.
INFO: Build completed successfully, 1606 total actions
➜  mjolnir git:(axum-serves-static-files) ✗ ls -lh bazel-bin/backend/prod
-r-xr-xr-x@ 1 Luke  staff   4.9M Mar 27 08:21 bazel-bin/backend/prod
```

## 3. Building release version with flags on BOTH binary and library (2.4 MB)
```
➜  mjolnir git:(axum-serves-static-files) ✗ bazel build --config=release //backend:prod
WARNING: Build options --@@rules_rust+//rust/settings:lto and --compilation_mode have changed, discarding analysis cache (this can be expensive, see https://bazel.build/advanced/performance/iteration-speed).
INFO: Analyzed target //backend:prod (664 packages loaded, 14757 targets configured).
INFO: Found 1 target...
Target //backend:prod up-to-date:
  bazel-bin/backend/prod
INFO: Elapsed time: 7.206s, Critical Path: 5.49s
INFO: 5 processes: 2304 action cache hit, 3 internal, 2 darwin-sandbox.
INFO: Build completed successfully, 5 total actions
➜  mjolnir git:(axum-serves-static-files) ✗ ls -lh bazel-bin/backend/prod              
-r-xr-xr-x@ 1 Luke  staff   2.4M Mar 27 08:21 bazel-bin/backend/prod
```

Since (3) has 2.4 MB and (1) has 2.5 MB, it seems that while there was indeed some optimization and reduction in size, it wasn't nearly as strong as the 4 something MB drop I said in the previous comment. It's true that in the previous one I hadn't yet embedded the frontend static assets into the binary, but I'm not sure what I did wrong previously. (UPDATE: jk I think I found my source of confusion, I'll add to the [following comment](https://github.com/luke-venk/mjolnir/pull/15#discussion_r3001263206))

I'm no longer sure keeping the flags on the library is justified, so I'll remove them, and just keep `Ctrip=debuginfo` on the library since you asked. Sorry about the confusion.

backend/BUILD.bazel
luke-venk: Good call. Nope, previously when deploy/BUILD.bazel was a thing it relied on the backend:prod binary being public, and for some reason I thought the library also had to be public, which was never true.

frontend/BUILD.bazel
luke-venk: This one is like the only target that needs to be public. The backend prod target relies on the frontend's static export build.

backend/BUILD.bazel
luke-venk: Here's what tripped me up. I think I mistook the switch from dev to prod binary sizes for optimizations. Here's a summary of commands vs binary sizes:

| Type | Optimized? | Binary Size |
|---|---|---|
| Dev | No | 5.4 MB |
| Dev | Yes | 1.9 MB |
| Prod | No | 4.9 MB |
| Prod | Yes | 2.5 MB |

So I think that explains where I got my ~4 MB drop "evidence", because when I was investigating binary size in the [previous PR comment](https://github.com/luke-venk/mjolnir/pull/13#discussion_r2972686931) I was working with a drop from 5.4 MB to 1.9 MB. So when I found my "new" drop was from 4.9 MB to 2.5 MB for prod, I mistakenly assumed it was an issue with the flags going on the library. At the same time I implemented a split into a common rust_lib and 2 rust_binaries, so all these changing things just confused me and I lost track of what really optimized what.

owens3364: There’s a bit of optimization called LTO optimization. I don’t remember what exactly that stands for but basically the compiler analyzes execution paths and removes dead code and unused bits and repeatedly optimizes/simplifies things until a very small binary with 100% actual usage is achieved. This is a “fat” LTO, a very time consuming and expensive task. “Thin” LTO is a much faster way to do the same thing, but is slightly less effective. If you look in the bazelrc, that’s what we’re using. IMO that’s the right setting.

I suspect that, because thin lto isn’t as thorough as fat, running thin lto against the library AND the binary is slightly more effective than thin lto against the binary only, which explains the 2.4 vs 2.5 MB difference. I would guess that running fat lto against the binary AND the library would result in no difference in binary sizes (ie binary with fat lto would be same size as binary with fat lto and library with fat lto).

owens3364 (approved):

PR #16, opened 2026-03-31
Added Simulation Logic to Backend & Add Real vs. Fake Configurations

luke-venk:
I added what you mentioned about using `cfg` to separate real vs. fake configurations. I also updated frontend accordingly to either run sims locally or use backend run sims.

I ran with `bazel run //backend:<dev/prod>_<real/fake>` and either with embedded frontend or Next.js running alongside it, it seems to work.

Some things to consider:
* When I run the simulations sometimes the javelin will appear to land in the throwing region or out of bounds, but not show an infraction. I thought I copied the math correctly but maybe you can clearly see if I didn't.
* The idea is that when the frontend Next.js server is being run by itself, it's build target gives an env var that tells it to run simulations locally (what you wrote). Otherwise, it sends an HTTP request to the backend to run the simulation.
* If the program is run in `real` mode, should I disable simulations (using cfg flags to disable routes)? Right now simulations can be run in any mode, but only in `real` mode do the CV pipelines start.


backend/src/server/server.rs
```diff
@@ -67,7 +61,8 @@ pub fn create_api_router() -> Router {
     // Define HTTP routes.
     let http_routes = Router::new()
         .route("/health", get(health_check))
-        .route("/throw-type", post(post_throw_type).get(get_throw_type));
+        .route("/throw-type", post(post_throw_type).get(get_throw_type))
+        .route("/simulate-throw", get(get_simulated_throw));
```
owens3364: This should just be a regular `/analyze-throw` endpoint, right? The frontend shouldn't have any idea if the incoming data is simulated or not, no?

backend/src/throws/simulate_throw.rs
```diff
@@ -0,0 +1,112 @@
...
+use std::f32::consts::PI;
+use uuid::Uuid;
+
+use super::{Infraction, InfractionType, ThrowType};
+
+// Refer to the `throwEventSchema` in frontend/lib/schemas.ts.
+// Allow camelCase in frontend and snake_case in backend.
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct GetSimulatedThrowResponse {
```
owens3364: I would call this `ThrowAnalysis` or `ThrowAnalysisResponse`

backend/src/throws/simulate_throw.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub throw_id: Uuid,
+    pub timestamp: String,
```
owens3364: Is this the timestamp of the throw start or throw end? Timestamp of request sending or response formulation? What timestamp is this? Do we need it?

backend/src/throws/simulate_throw.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub throw_type: ThrowType,
+    pub distance: f32,
```
owens3364: Add a unit (meters probably)

backend/src/throws/simulate_throw.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub infractions: Vec<Infraction>,
+    pub images: Vec<String>,
+
+    // Note: Currently frontend doesn't return a landing point if there is
+    // an infraction, but shouldn't it still return a landing point if
+    // there is a circle infraction and not a sector violation?
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub landing_point: Option<(f32, f32)>,
```
owens3364: Specify if in x/y or polar coords... I think we typically use xy but we should be specific bc Ik we have some polar coordinate math

backend/src/throws/simulate_throw.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+pub fn get_circle_field_dims(throw_type: ThrowType) -> (f32, f32) {
```
owens3364: Include units and specify what these actually mean - this is throwing arc radius and max distance, I think, but it doesn't include sector angles so those must be elsewhere... idk just kind of confusing way to implement dimensions.

It is super possible that you translated my typescript code and it turned into this, ie that it's my fault for shitty organization... but I'm speaking up now!

backend/src/throws/simulate_throw.rs
```diff
@@ -0,0 +1,112 @@
...
+        ThrowType::Discus => (2.5, 80.0),
+        ThrowType::Hammer => (2.135, 90.0),
+        ThrowType::Javelin => (16.0, 100.0),
+    }
+}
+
+pub fn get_random_infractions() -> Vec<Infraction> {
+    let mut infractions: Vec<Infraction> = vec![];
+
+    if rand::random::<f32>() < 0.3 {
```
owens3364: Nit: could extract 0.3 into a const variable with a name instead of reusing 3x

backend/src/throws/simulate_throw.rs
owens3364: lmao yep it's on me

backend/BUILD.bazel
```diff
@@ -29,13 +32,17 @@ rust_library(
 # the frontend. Optimization can technically occur for either target
 # using `--config=release` (see .bazelrc).
 
+# TODO: add documentation for fake vs. real.
```
owens3364: Now would be a good time :)

backend/BUILD.bazel
```diff
@@ -29,13 +32,17 @@ rust_library(
...
+
 # Dev mode for the backend doesn't have to be optimized. In this mode,
 # the backend can serve the API via command line and it runs alongside
 # the Next.js server. Since the Next.js requests come from a different
 # origin, so CORS is needed. It will not have any embedded assets.
```
owens3364: Nit: remove 'so'

frontend/BUILD.bazel
```diff
@@ -86,6 +87,7 @@ js_run_devserver(
     # Since this is in dev mode, don't static export the frontend.
     env = {
         "NEXT_PUBLIC_API_BASE_URL": "http://localhost:5001",
+        "NEXT_PUBLIC_FETCH_FROM_AXUM": "true",
```
owens3364: Question: Why do these vars start with `NEXT_PUBLIC` ? What is that supposed to indicate? This could be the correct thing I just don't understand it.

Re `FETCH_FROM_AXUM` - should probably be called `USE_EXTERNAL_API` or something. Reason is, frontend shouldn't care what web server framework rust is using under the hood, it just needs to know to use the external API instead of an internal one

owens3364 (changes requested):

frontend/BUILD.bazel
luke-venk: Here's the documentation for that: https://nextjs.org/docs/pages/guides/environment-variables#bundling-environment-variables-for-the-browser

To summarize, without these prefixes, these env vars wouldn't be available in the browser

backend/src/throws/simulate_throw.rs
luke-venk: Currently it's just the timestamp of when the response is being formulated, in both this file and in `frontend/lib/api.ts`. I'm not sure how useful this is, I was just matching behavior in Next.js.

In practice, I guess we'd need timestamps associated with the actual sport, right? Things like exact time of ground impact, circle exit, etc.?

backend/src/throws/simulate_throw.rs
luke-venk: Updated both this and throwEventSchema's `distance` field in `frontend/lib/schemas.ts`

backend/src/throws/simulate_throw.rs
luke-venk: I specified xy and updated frontend.

As the comment says, should this be updated to still give a landing point even if there is a circle infraction, as long as there is no sector violation?

backend/src/throws/simulate_throw.rs
luke-venk: I added a docstring explaining it's diameter of circle and length of field. I'll also add a third return for sector angles, but in Next.js it's hardcoded within the `getThrowEvent()` function right now. I can update it if you'd like.

Also, why is the javelin "diameter" field 16m instead of 30m? I'm asking per this link saying "The runway is 30m minimum long" for javelin.
https://www.cits.wa.gov.au/sport-and-recreation/sports-dimensions-guide/athletics-throwing-events

owens3364: The diameter is actually the arc diameter of whatever circular boundary

So even though there's a runway, our field coordinates still treat 0 0 as the center of the throwing circle, and the arc of the javelin throw setup has a radius of 8m (diameter of 16m).

So that bit is correct.

backend/src/throws/simulate_throw.rs
owens3364: I think it should be timestamp in form of microseconds since unix epoch based on the reference provided by the timestamps on the PTP-synced frames from the cameras. So yash's code should return landing distance and that timestamp. It's ok if we fake the value right now, just saying the timestamp variable name and comments should be updated to make this the clear timestamp scheme.

backend/src/throws/simulate_throw.rs
owens3364: yep!

backend/src/throws/simulate_throw.rs
owens3364: Don't worry about updating the Next.js, I'll make a follow up PR that follows this one to improve the website's "real"ness

frontend/BUILD.bazel
owens3364: Great to know! Did not realize that. Cool!

backend/src/throws/simulate_throw.rs
luke-venk: I'm down for that, but right now the frames in `backend/src/camera/record/capture.rs` have the following in the metadata JSON files.

```
let system_timestamp_ns = buffer.system_timestamp();
let buffer_timestamp_ns = buffer.timestamp();
```

The docstring for `system_timestamp()` says "Gets the system timestamp for when the frame was received. Expressed in nanoseconds."

The docstring for `timestamp()` says "Gets the buffer camera timestamp, expressed as nanoseconds. Not all devices provide reliable timestamp, which means sometimes its better to rely on the buffer completion host local time, or to use `system_timestamp()`".

So do we want to make the timestamp here in nanoseconds instead of microseconds? Also it sounds like we'd care about `system_timestamp()`, it's probably time since Unix Epoch, right?

backend/src/throws/throw_analysis.rs
```diff
@@ -8,15 +8,27 @@ use uuid::Uuid;
 #[serde(rename_all = "camelCase")]
 pub struct ThrowAnalysisResponse {
     pub throw_id: Uuid,
-    pub timestamp: String,
+
+    // The time since Unix epoch, in nanoseconds, at the time the last
+    // frame of impact was captured.
+    pub timestamp_final_frame_ns: String,
```
owens3364: Oooh I don't like this as a String but I understand why... I can't think of a better solution we'll just have to keep in mind that it's a string on frontend

luke-venk: Why does that matter, is there a specific datetime type that's better? I think in Next.js it's already a string, could be mistaken though

```
timestampFinalFrameNs: z.string().refine((val) => !isNaN(Date.parse(val)), {
      message: "Invalid timestamp",
}),
```
snippet in `frontend/lib/api.ts`

EDIT: Also you may have been alluding to this but timestamps are string rn yeah, but if we want it to be the buffer's timestamp that would be an unsigned integer, right?

backend/src/throws/simulate_throw.rs
owens3364: Mmm the "not all devices provide reliable timestamp"... is ambiguous because aravis has to support a bunch of cameras, but for our specific setup with PTP synced cameras, those timestamps are the ones we want to reference by. System timestamp is the timestamp recorded by the computer when it received the frame from the camera, which is going to be way off and totally different from the actual moment the frame was captured.

Since yash needs relative time differences and not absolute times, it's fine if the camera timestamp isn't perfectly synced up with the unix epoch.

I do expect that the cameras will be synced to the unix epoch though via ntp or something like that without us even having to think about it.

Regardless, use `.timestamp()` and not `.system_timestamp()`

Re ns vs microseconds... I would go for microseconds. We don't need the nanoseconds precision, nor does it actually exist, and in the event some code somewhere like multiplies times or computes an average or something, microseconds will have lower overflow risk.

So maybe call it `frame_timestamp_from_camera_microseconds` ?

luke-venk: Sounds good, I'll do that. Tagging @yashj1579 so he sees this. I'll also add this comment to this ticket so the information is there: #28

owens3364 (approved):

backend/src/throws/throw_analysis.rs
owens3364: Yeah I am alluding to the fact that dates should be numbers or objects around numbers. Dates as strings kind of suck- string encoding, more data usage, no native math or comparison (has to be parsed to number or bigint first), possibility of blank strings or bad or malformed chars… ew.

that said, the JS number type is quite dogshit. JS numbers are 64-bit floats, no exceptions. If you are representing an int, you can do a 53-bit int, but not higher without it becoming an inexact 64-bit float approximation of an integer. So then you have to think about “have there been more than 2^53 nanoseconds since the Unix epoch” which is kind of ridiculous and make sure you have enough margin for computations like averages or whatever.

finally, the transmission medium is JSON, and json has no support for bigint. So you’re stuck with JS number instead of bigint, which would solve this problem, and you fallback on string because it’s the only thing that can definitely represent your date number without screwing up the value, and then you can convert to bigint or whatever.

PR #25, opened 2026-04-09
Add Tools for Streaming and Recording with Cameras

luke-venk:
Features are captured in the commit names.


backend/src/bin/record_from_both_cameras.rs
```diff
@@ -0,0 +1,71 @@
+/// Tool for users to record footage from both cameras using Aravis and
+/// store the frames to disk using the command-line.
+use std::thread;
+
```
owens3364: Nit: rm newline

backend/src/bin/record_from_both_cameras.rs
```diff
@@ -0,0 +1,71 @@
...
+use backend_lib::camera::aravis_utils::initialize_aravis;
+use backend_lib::camera::CameraIngestConfig;
+use backend_lib::camera::discovery::get_camera_ids;
+use backend_lib::camera::record::cli::RecordWithBothCamerasArgs;
+use backend_lib::camera::record::record_from_one_camera;
+use backend_lib::camera::record::writer::{ensure_dir, string_to_pathbuf};
+
```
owens3364: Nit: rm newline

backend/src/camera/record/capture.rs
```diff
@@ -0,0 +1,155 @@
...
+                    "Frame {} received at {:.2}s",
+                    frames_saved,
+                    elapsed_since_start.as_secs_f64()
+                );
+
+                let data = copy_buffer_bytes(&buffer);
+
+                if data.is_empty() {
+                    eprintln!("Empty buffer from camera {}.", config.camera_id);
+                    stream.push_buffer(buffer);
```
owens3364: We should push the buffer back to the stream as soon as we call `copy_buffer_bytes` so the camera stream doesn't starve.

backend/src/camera/record/writer.rs
```diff
@@ -0,0 +1,89 @@
...
+use serde_json::to_string_pretty;
+
+/// Helper function to ensure output directory exists.
+pub fn ensure_dir(path: &PathBuf) {
+    fs::create_dir_all(path)
+        .unwrap_or_else(|e| panic!("Failed to create directory {}: {e}", path.display()));
+}
+
+/// Helper function to create output directory.
+pub fn string_to_pathbuf(path: &String) -> PathBuf {
```
owens3364: IMO remove this function... it isn't making your code any shorter and obscures a standard library function.

backend/src/camera/record/capture.rs
```diff
@@ -0,0 +1,155 @@
...
+                    payload_bytes: data.len(),
+                    system_timestamp_ns: buffer.system_timestamp(),
+                    buffer_timestamp_ns: buffer.timestamp(),
+                    frame_id: buffer.frame_id(),
+                    exposure_time_us: config.exposure_time_us,
+                    frame_rate_hz: config.frame_rate_hz,
+                };
+
+                // Write the frame's raw bytes and metadata to file.
+                write_frame_files(
```
owens3364: This is probably fine if we have bad framerates, but I would *strongly* consider making a single filewriter async tokio task and using a bounded crossbeam channel to pipe frame data to it.
If your disk latency is a problem at all it will fuck your camera stream.

You can make a single crossbeam channel and clone the tx so that you have 2 tx objects, and pass one to each camera capture thread. Then each camera thread can send frames over tx, and your filewriter thread / async task will receive both sets of frames to write on one rx object. Pretty neat!

backend/src/camera/stream/capture.rs
owens3364, line 1: Ok I see a stream capture.rs and a record capture.rs ... there looks to be a lot of code duplication. Is there a reason we have two very similar things?

If the answer is that we're in a rush and we can refactor later, that's fine, I just want to make sure we're clear about which is used for what and their differences because this would be an easy way for bugs to slip into one or the other.

backend/src/camera/aravis_utils.rs
```diff
@@ -0,0 +1,131 @@
...
+    // Use binning to downsample (if necessary) from full resolution
+    // to smaller resolution.
+    camera
+        .set_region(0, 0, 4096, 3000)
+        .expect("Failed to set resolution in camera configuration.");
+    if camera.is_binning_available().expect("Error: Binning is not available for this camera.") {
+        camera.set_binning(config.resolution.binning(), config.resolution.binning()).expect("Error: Failed to set binning for camera.");
+    }
+
+    // Aperture.
```
owens3364: Delete commented out code if we ain't gon use it
In our case we won't use this, I had to look it up but the camera aperture is a lens-dictated setting

backend/src/camera/aravis_utils.rs
```diff
@@ -0,0 +1,131 @@
...
+    // Packet size.
+    // GigE Vision streams camera data over UDP. Automatically setting the packet
+    // size works by sending test packets of decreasing sizes until one gets through
+    // without fragmentation, discovering the Maximum Transmission Unit (MTU), the
+    // largest payload that can travel end-to-end without being split.
+    // Standard Ethernet MTU is about 1500 bytes, but for jumob packets we need
+    // around 8064 bytes. For some reason, automatic packet size negotation isn't
+    // detecting an MTU of higher than 1508, so we configure it manually here.
+
+    // camera
```
owens3364: Delete

backend/src/camera/aravis_utils.rs
```diff
@@ -0,0 +1,131 @@
...
+    //     .gv_auto_packet_size()
+    //     .expect("Failed to automatically set the packet size in camera configuration");
+
+    // println!(
```
owens3364: Delete

backend/BUILD.bazel
```diff
@@ -84,3 +91,45 @@ rust_test(
...
+rust_binary(
+    name = "discover",
+    srcs = ["src/bin/discover_cameras.rs"],
+    deps = [
+        ":backend_lib",
+    ]
+)
+
+rust_binary(
+    name = "record-from-one",
```
owens3364: Nit: use underscores
Almost everything else in bazel uses underscores, so just easier to keep ppl in the habit :)

owens3364 (changes requested):

backend/src/camera/stream/capture.rs
luke-venk: You're right there's some code duplication, but there are also quite a few differences:

| `record/capture.rs` | `stream/capture.rs` |
|-|-|
| * Writes raw bytes to disk and uses `writer.rs` modules | * Shares camera ingest config with UI thread via Arc<Mutex<T>> |
| * Keeps track of metadata to write in JSON file | * Supports restarting stream if user requests it |
| * Keeps track of stopping conditions (max frame or max duration) | * Sends frames to UI thread via crossbeam channel |

If you think it would be cleaner to refactor, I can definitely do that another time.

backend/src/camera/record/capture.rs
luke-venk: Good idea. I implemented the separate threads, but I did so without tokio since we didn't use tokio for streaming. Is tokio necessary?

owens3364: Oh yeah no worries re no tokio - you could approach it with definite threads or async tasks, both should work well here; your approach is actually probably better!

backend/src/camera/stream/capture.rs
owens3364: Another time sounds good :)

owens3364 (changes requested):
Ok this is pretty dope.
Only thing I would suggest is that so far every binary has to be killed coarsely by Ctrl + C, which kills the program without stopping acquisition or doing any communication with the cameras or aravis, which is pretty rough. We know this is unstable because sometimes we try to restart recording or streaming or whatever and get errors on startup and have to Ctrl C and restart.

I found a rust crate that does this https://docs.rs/ctrlc/latest/ctrlc/ and I don't think it should be too bad to add a handler that stops each task and stops acquisition.

If you want to kick this to another PR that would be ok too, but I do think it would be helpful for more camera performance stability during recording sesh today.

owens3364 (approved):

PR #26, opened 2026-04-09
Loseless Compression

grao23:
The code adds lossless H.265 compression to camera recording. When --compress is enabled, frames from each camera are no longer written as raw per-frame dumps; instead, each camera’s frames are fed into one long-lived ffmpeg encoder that writes a single lossless .h265 stream for that recording session. The recording flow is controlled through the shared recording CLI/config path, the compression logic is organized under backend/src/camera/record/compression, and the recovery step can decode a saved .h265 stream back into PNG frames when --recover-to-png-dir is provided.


backend/src/bin/record_from_both_cameras.rs
```diff
@@ -18,18 +18,31 @@ pub fn main() {
...
-    args.common_args.validate().unwrap_or_else(|err| panic!("{err}"));
+    args.common_args
+        .validate()
+        .unwrap_or_else(|err| panic!("{err}"));
 
     // Create output directory based on command-line argument.
     let output_base_dir = string_to_pathbuf(&args.common_args.output_dir);
     ensure_dir(&output_base_dir);
 
+    let recover_base_dir = args
```
luke-venk, line 29: This line should just look like how `output_base_dir` is handled right above, unless I'm missing something

backend/src/bin/record_from_one_camera.rs
```diff
@@ -14,22 +15,36 @@ pub fn main() {
...
-    args.common_args.validate().unwrap_or_else(|err| panic!("{err}"));
+    args.common_args
+        .validate()
+        .unwrap_or_else(|err| panic!("{err}"));
 
     // Create output directory based on command-line argument.
     let output_base_dir = string_to_pathbuf(&args.common_args.output_dir);
     ensure_dir(&output_base_dir);
 
+    let recover_base_dir = args
```
luke-venk, line 26: Same as https://github.com/luke-venk/mjolnir/pull/26/changes#r3118744625

backend/src/bin/recover_h265_to_png.rs
```diff
@@ -0,0 +1,45 @@
+use std::path::PathBuf;
+
```
luke-venk, line 2: nit: get rid of empty line and run formatter

MODULE.bazel
```diff
@@ -54,6 +54,10 @@ crate = use_extension(
 )
 
 # Add external Rust dependencies here.
+crate.spec(
+    package = "anyhow",
```
luke-venk, line 58: We haven't used `anyhow` for error handling in other parts of the project. Is there a specific reason you are introducing it now?

backend/src/camera/record/writer.rs
```diff
@@ -1,12 +1,12 @@
-/// Code that handles writing our captured frames from RAM to disk (SSD) in
-/// a performant manner so frames aren't dropped.
+/// Code that handles writing our captured frames from RAM to disk in
```
luke-venk, line 1: Why was this comment changed to no longer make grammatical sense lol

backend/src/camera/record/writer.rs
```diff
@@ -41,40 +41,69 @@ pub fn sanitize_path_name(value: &str) -> String {
         .collect()
 }
 
-/// Writes the captured frame and metadata to disk.
+/// Paths written for one saved frame.
+pub struct WrittenFramePaths {
```
luke-venk, line 45: I'm not sure this is needed. Can you please explain why you added this?

backend/src/camera/record/compression/frames.rs
```diff
@@ -0,0 +1,161 @@
+// Compresses per-frame MONO8 recordings with zstd and recovers them to PNG.
```
luke-venk, line 1: If `frames.rs` is the only file in `compression/`, can you just make it a `compression.rs` file and move it back into `record/` instead of having a separate directory with only 1 file?

backend/src/camera/record/compression/frames.rs
```diff
@@ -0,0 +1,161 @@
...
+use anyhow::{Context, Result, bail};
+use png::{BitDepth, ColorType, Encoder as PngEncoder};
+
+use crate::camera::record::writer::FrameMetadata;
+
+pub const COMPRESSED_FRAME_EXTENSION: &str = "zst";
+const ZSTD_LEVEL: i32 = 1;
+
+// Compresses one raw MONO8 frame with a fast zstd setting.
+pub fn compress_mono8_frame(raw_bytes: &[u8]) -> Result<Vec<u8>> {
```
luke-venk, line 15: How do we know that specifically H.265 compression is used? I tried to search for things like "h265" and couldn't find that in our code anywhere

backend/src/camera/record/compression/frames.rs
```diff
@@ -0,0 +1,161 @@
...
+        let height =
+            u32::try_from(metadata.height).context("metadata height does not fit into u32")?;
+        write_mono8_png(&png_path, width, height, &raw_bytes)?;
+    }
+
+    Ok(frame_paths.len() as u64)
+}
+
+#[cfg(test)]
+mod tests {
```
luke-venk, line 102: Correct me if you've already done this, but wouldn't it be extremely useful if we had a unit test for checking that if you pass in a frame and test the following
- Compression doesn't fail, and the result is smaller in size than the input
- Uncompression doesn't fail, and the result is EXACTLY THE SAME as the original input
?

backend/src/camera/record/capture.rs
```diff
@@ -146,6 +168,18 @@ pub fn record_from_one_camera(
     }
 
     let _ = camera.stop_acquisition();
+    if config.compress {
+        if let Some(recover_camera_dir) = recover_camera_dir.as_ref() {
+            let recovered = recover_compressed_dir_to_pngs(&output_camera_dir, recover_camera_dir)
```
luke-venk, line 173: This seems to automatically kick off PNG conversion right after we kill our program, right? Do we want this to happen always?

I'd imagine conversion to PNG takes a lot of time and we might not want this always to kick off. And if we did have it always kick off, what's the point of the binary `recover_h265_to_png.rs`?

luke-venk (changes requested):
The `dev/discover-and-record-from-cameras` branch no longer exists, so this PR should be updated to be merged into `main`

After that, there very likely be merge conflicts

PR #27, opened 2026-04-11
Circle infraction system ingest

owens3364:


MODULE.bazel
```diff
@@ -55,9 +55,9 @@ crate = use_extension(
 
 # Add external Rust dependencies here.
 crate.spec(
+    default_features = False,
```
luke-venk, line 58: Is there a standard somewhere where features should be above package and version, or is that your preference?

backend/src/server/server.rs
```diff
@@ -1,4 +1,6 @@
 use crate::{schemas::ThrowType, server::app_state::AppState};
```
luke-venk: nit: run formatter so imports cleaner

backend/src/server/server.rs
```diff
@@ -60,9 +62,26 @@ async fn get_throw_type(State(state): State<AppState>) -> Json<GetThrowTypeRespo
 
 // In both dev and prod mode, the router will require the HTTP routes
 // and the thread-safe shared app state.
-pub fn create_api_router() -> Router {
+pub fn create_api_router(circle_rx: Receiver<CircleInfractionDetectionState>) -> Router {
     // Thread-safe shared app state for current throw event.
     let state = AppState::new();
+    let state_clone = state.clone();
+    tokio::spawn(async move {
```
luke-venk, line 75: Question, why use Tokio instead of just normal spawning a thread? I looked it up and it seems to be better for I/O, and the Axum server is already built on Tokio anyways.

backend/src/server/server.rs
luke-venk: Can you update unit tests to use your new `create_api_router` signature?

backend/src/circle_infractions_ingest/infraction_byte_decoder.rs
```diff
@@ -0,0 +1,52 @@
...
+/// Returns None if the byte is too ambiguous to correct.
+pub fn decode(raw: u8) -> Option<InfractionState> {
+    if raw == InfractionState::Clear as u8 {
+        return Some(InfractionState::Clear);
+    }
+    if raw == InfractionState::Infraction as u8 {
+        return Some(InfractionState::Infraction);
+    }
+
+    let d_clear = hamming(raw, InfractionState::Clear as u8);
```
luke-venk, line 26: Very neat. I've never worked with hamming distance before in a project

backend/src/circle_infractions_ingest/infraction_byte_decoder.rs
```diff
@@ -0,0 +1,52 @@
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum InfractionState {
+    Clear = 0x00,      // 0000_0000
+    Infraction = 0xFE, // 1111_1110
```
luke-venk, line 4: Very clever! Out of curiosity, why not make Infraction 0xFF? Are we using that last bit for something?

backend/src/circle_infractions_ingest/circle_infractions_thread.rs
```diff
@@ -0,0 +1,133 @@
...
+use std::time::Duration;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum CircleInfractionDetectionState {
+    Stale,
+    DetectedInfraction(u64), // timestamp of when infraction was detected in milliseconds since unix epoch
+    KeepAlive,
+}
+
+const STALE_THRESHOLD: f64 = 5.0; // Hz
```
luke-venk: nit: add "_HZ" to variable name?

backend/src/server/server.rs
```diff
@@ -60,9 +62,26 @@ async fn get_throw_type(State(state): State<AppState>) -> Json<GetThrowTypeRespo
 
 // In both dev and prod mode, the router will require the HTTP routes
 // and the thread-safe shared app state.
-pub fn create_api_router() -> Router {
+pub fn create_api_router(circle_rx: Receiver<CircleInfractionDetectionState>) -> Router {
```
luke-venk: Note: will have to update `create_api_router` usage in the `main.dev.rs` and `main.prod.rs` files

backend/src/circle_infractions_ingest/circle_infractions_thread.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+}
+
+/// Finds the arduino on a serial port and begins reading in data
+/// Returns a channel of CircleInfractionDetectionState updates, which will be
+/// - Stale if we haven't heard from the arduino in a 5 cycles (assuming 20Hz cycles)
+/// - DetectedInfraction when we detect an infraction
+/// - KeepAlive so we can detect a dropped channel and exit the thread gracefully
+/// Panics if the channel is full, which would indicate that the receiving end is not doing its job
+pub fn begin_detecting_circle_infractions(baud: u32) -> Receiver<CircleInfractionDetectionState> {
```
luke-venk, line 66: Looks good, would this just be called in the `main()` functions in `main.dev.rs` and `main.prod.rs`?

luke-venk (changes requested):

backend/src/circle_infractions_ingest/infraction_byte_decoder.rs
luke-venk: Also, tagging @EloghosaE so she can see this

MODULE.bazel
owens3364: Oh, changes to this file are in part the result of my autoformatter running. I have the Bazel VS Code extension and that includes an autoformatter for bazel build files. So I guess it is the bazel team's preference!

backend/src/circle_infractions_ingest/infraction_byte_decoder.rs
owens3364: 0xFE is actually a 4-bit 1111 and a 4-bit CRC 1110 (depending on the polynomial you choose ofc). But because our data is so well-defined we don't have any runtime CRC logic, which is great!
I think 0xFE is nice because you know you didn't get that just because the line was pulled high forever (hardware bug) and by that logic I should change 0x00 to 0x01 so I'll do that rq

backend/src/circle_infractions_ingest/circle_infractions_thread.rs
owens3364: Yep!

backend/src/server/server.rs
owens3364: You right

backend/src/server/server.rs
owens3364: In this case the threading isn't about hardware IO, it's about receiving from another crossbeam channel, and tokio is more suited to that because
- Tokio tasks are much more lightweight than OS threads
- Tokio is smart enough to yield this task for other things while waiting for a message from crossbeam


You are correct that for long-running IO-intensive things, OS threads are better (camera ingest is a prime example) but for things where the computation and IO are light but IO blocking dominates the execution time, then it makes much more sense to have a lot of async tasks that all yield when there is blocking IO and are all run by the same couple of worker threads (this is exactly what the tokio runtime provides, and exactly what axum uses under the hood).

Imagine being axum
- You are supposed to scale to thousands or millions of requests per second
- Server request handling is almost always IO-bound
    - The server computes the request response, serializes the exact bytes to send back over the network, copies them to the heap if they weren't there already, and then makes an OS syscall to have the OS copy that memory into the network card and/or DMA buffer attached to the network interface. The most latency intensive operation is waiting for the OS to copy out that memory and return from the syscall, even though almost no compute is required for that.
- So, you treat every single incoming request as a new tokio task, spin up as many worker threads as you have CPUs, and every task serializes a response and yields its execution time back to the other tasks once the OS syscall has been made. Once the OS syscall finishes, the tokio task for that request is completed and tasks from new requests take its place. But all the other tasks/worker threads can get response serialization done while the OS syscall blocks.
- This pattern is exactly how server frameworks like Node.js, gin/martini (golang), Ktor (Kotlin), and now Axum all work and are able to scale to dozens of thousands of requests per second despite many of those languages being single-thread (js), or no shared memory (golang iirc), or very expensive threading (kotlin / jvm threads are pretty heavy iirc but maybe they've fixed that)

luke-venk: Dude thanks for typing all that out. It makes sense. My basic 20 word understanding is:

std::thread spawns an actual OS thread, good for CPU-bound work. Tokio thread are tasks not 1-1 mapping with OS threads, instead scheduled by Tokio runtime, much lighter weight and good for I/O where you wait around for stuff

backend/src/bin/live_stream.rs
```diff
@@ -1,3 +1,10 @@
+use backend_lib::camera::CameraIngestConfig;
+use backend_lib::camera::stream::capture::run_capture_thread;
+use backend_lib::camera::stream::cli::StreamFromCamerasArgs;
+use backend_lib::camera::stream::{FrameData, LiveViewApp};
+use clap::Parser;
+use eframe::egui;
+use std::sync::atomic::{AtomicBool, Ordering};
 /// Tool for users to stream footage from the cameras using Aravis and
```
luke-venk, line 8: Idk why autoformatter always messes up comments but whatever

EDIT: lmao I looked it up and it's because `///` comments just attach to whatever it's directly above. So when format runs, it sorts all the actual lines of code and just moves the comments above it smh. This wouldn't have happened if I made the comments `//`

Whatever lmao

backend/src/circle_infractions_ingest/circle_infractions_thread.rs
```diff
@@ -0,0 +1,173 @@
...
+    thread::spawn(move || {
+        loop {
+            if let Err(TrySendError::Disconnected(_)) =
+                tx.try_send(CircleInfractionDetectionState::KeepAlive)
+            {
+                return;
+            }
+            thread::sleep(Duration::from_millis(5));
+            if rand::random::<f64>() < 1.0 / 1200.0 {
+                // simulate ~1 infraction per minute
```
luke-venk, line 155: niceee

luke-venk (approved):

backend/src/server/server.rs
owens3364: Yes this is correct!

backend/src/bin/live_stream.rs
owens3364: Yes /// is rustdoc so you can see in editor while // is more for non-doc comments and does not appear in editor

PR #29, opened 2026-04-12
adding latest cv code 

rx9933:
pipeline.py is an older version without: consistency-based roi (if N identified points, use ROI to select candidates), candidate score (based on circularity, area, and other parameters for shotput, etc. I tuned the CV algorithm for the data collected on 04/09.

### DIFF
--- a/analysis/opencv_approach/params.py
+++ b/analysis/opencv_approach/params.py
@@ -1,0 +1,57 @@
+# 1k
+# (1024, 750)
+
+# (2048, 1500),
+
+# 4k:
+# (4096, 3000),
+
+p = '4k'
+SCALE = 1.0
+DISPLAY_WIDTH, DISPLAY_HEIGHT  = 1920, 1080
+PROCESS_HEIGHT, PROCESS_WIDTH = 960, 540
+
+
+if p == '4k':
+    SCALE = 4.0
+    MOG2_VAR_THRESHOLD  = 80    # lower = more sensitive; raise if noisy background
+elif p == '2k':
+    SCALE = 2.0
+
+
+AREA_SCALE = SCALE ** 2
+
+def s(x): return int(x * SCALE)
+def a(x): return int(x * AREA_SCALE)
+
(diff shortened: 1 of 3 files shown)


owens3364 (changes requested):
It would be super helpful if we could see the diff between this and the pipeline.

Could you either
- Modify the existing pipeline file
OR
- copy the existing pipeline file, merge it, and then submit a PR with your changes to the copy? This would keep the different versions still in the codebase.

rx9933:
I think this is what you wanted -- pipeline.py now shows the updated difference. Sorry for the delay.

owens3364 (changes requested):
I don't mean this in an insulting way, but compared to main, I see about a 20% performance hit by fps in `tuned_cv` in comparison with `pipeline` on main when both are run against the same video. Additionally, I see `tuned_cv` appearing to be less reliable in maintaining a lock on the object during its trajectory. Also, `pipeline` is completely unusable on this branch (unless I'm doing something wrong when running it).

Am I missing something? Is there a goal accomplished that I'm not seeing? I know you spent time on this and I'm not tryna dog on your efforts, I just am not seeing why these changes should be merged.

rx9933:
Hi Owen,

- Are you running the two codes with the latest data (e.g., from the last field day or the last two field days)? Also, can you try the `tuned_cv `with the 1k setting in params? That should speed up the compute. 
- I believe you are right that `tuned_cv` is a bit slower than `pipeline`, but IMO it is more accurate (also, Yash said he was getting better results with `tuned_cv` rather than pipeline, a couple weeks ago). For reference, please view the cv_comparision doc below.
- Also, in both cases, the longer videos (for background subtraction) do not seem to help the accuracy significantly. For reference, I have attached the latest mp4 videos on [UT Box: ](https://utexas.box.com/s/nwhmif56sx9ioo3d2ixu7sgvoy4nrnxu)


[cv_comparision.docx](https://github.com/user-attachments/files/27248531/cv_comparision.docx)

Issue #30, opened 2026-04-12
Fix Optimized Build Bug Related to Eframe


luke-venk:
## Description
When building config releases with `bazel build --config=release //backend:prod-fake`, the build fails.
```
➜  mjolnir git:(main) bazel build --config=release //backend:prod-fake
WARNING: Build options --@@rules_rust+//rust/settings:lto and --compilation_mode have changed, discarding analysis cache (this can be expensive, see https://bazel.build/advanced/performance/iteration-speed).
INFO: Analyzed target //backend:prod-fake (846 packages loaded, 23386 targets configured).
ERROR: /Users/Luke/Code/mjolnir/backend/BUILD.bazel:86:12: Compiling Rust bin prod-fake (1 file) failed: (Exit 1): process_wrapper failed: error executing Rustc command (from rust_binary rule target //backend:prod-fake) bazel-out/darwin_arm64-opt-exec/bin/external/rules_rust+/util/process_wrapper/process_wrapper --arg-file ... (remaining 349 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
error: the crate `objc2` requires panic strategy `unwind` which is incompatible with this crate's strategy of `abort`

error: the crate `objc2_foundation` requires panic strategy `unwind` which is incompatible with this crate's strategy of `abort`

error: the crate `block2` requires panic strategy `unwind` which is incompatible with this crate's strategy of `abort`

error: the crate `objc2_core_foundation` requires panic strategy `unwind` which is incompatible with this crate's strategy of `abort`

error: the crate `objc2_metal` requires panic strategy `unwind` which is incompatible with this crate's strategy of `abort`

error: the crate `objc2_app_kit` requires panic strategy `unwind` which is incompatible with this crate's strategy of `abort`

error: the crate `objc2_core_graphics` requires panic strategy `unwind` which is incompatible with this crate's strategy of `abort`

error: aborting due to 7 previous errors

Target //backend:prod-fake failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 44.278s, Critical Path: 30.61s
INFO: 2021 processes: 611 action cache hit, 1132 internal, 887 darwin-sandbox, 2 local.
ERROR: Build did NOT complete successfully
```

[...]

PR #38, opened 2026-04-14
Fix Optimization Bug Related to egui & eframe

luke-venk:
Related to #30

I also changed `///` comments to `//` at top of files per our convo yesterday and ran autoformatter. The only files you need to review are the ones changed in the first commit.


backend/BUILD.bazel
```diff
@@ -61,6 +67,29 @@ rust_library(
     ],
 )
 
+# Define a secondary library for streaming, which uses eframe and egui.
```
owens3364, line 74: Good comment I wondered about this

owens3364 (approved):

backend/BUILD.bazel
```diff
@@ -61,6 +67,29 @@ rust_library(
...
+# We do this because optimized release builds will use the `abort` panic
+# strategy to decrease binary size, but eframe and egui require the `unwind`
+# panic strategy. We would like to exclude this library from optimization
+# since it would throw an error and it's not needed in the final binary anyways.
+# See ticket #30 for more information.
+rust_library(
+    name = "stream_lib",
+    crate_name = "stream_lib",
```
owens3364: Nit: you don't need the explicit `crate_name` unless it is different from the regular `name` property

backend/BUILD.bazel
```diff
@@ -61,6 +67,29 @@ rust_library(
     ],
 )
 
+# Define a secondary library for streaming, which uses eframe and egui.
+# We do this because optimized release builds will use the `abort` panic
+# strategy to decrease binary size, but eframe and egui require the `unwind`
+# panic strategy. We would like to exclude this library from optimization
+# since it would throw an error and it's not needed in the final binary anyways.
+# See ticket #30 for more information.
+rust_library(
```
owens3364, line 80: Have we verified that this strategy actually results in the optimizations not being applied to this library? Ie, will this library, once used by a binary, inherit whatever optimization flags the binary has? If so, does this change still have the impact you wanted?

owens3364 (changes requested):

backend/BUILD.bazel
luke-venk: As is, there are no optimization flags on the `stream` binaries or libraries, so no issues yet. But you're right I should add optimization flags to all the auxiliary binaries since we want to ship the smallest possible binaries. I will do this now.

To answer your question, for most optimization flags, the library doesn't inherit from the binary. But the panic strategy flag IS inherited, the only flag for us that is like that. I guess it makes sense since if you wanted your shit to panic a certain way, you'd want ALL your code to panic the same way. See relevant docs [here](https://doc.rust-lang.org/rustc/codegen-options/index.html#panic):
> If any crate in the crate graph uses abort, the final binary (bin, dylib, cdylib, staticlib) must also use abort

So basically, since the optimized binary would ideally use the `abort` panic strategy, the library also would. But the library uses crates that use other crates that are incompatible with `abort`

```
mjolnir git:(fix-optimization-bug-egui-eframe) ✗ bazel build --config=release //backend:record                                                                                 
INFO: Analyzed target //backend:record (0 packages loaded, 0 targets configured).
ERROR: /Users/Luke/Code/mjolnir/backend/BUILD.bazel:254:12: Compiling Rust bin record (1 file) failed: (Exit 1): process_wrapper failed: error executing Rustc command (from rust_binary rule target //backend:record) bazel-out/darwin_arm64-opt-exec/bin/external/rules_rust+/util/process_wrapper/process_wrapper --arg-file ... (remaining 180 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
error: the crate `dispatch2` requires panic strategy `unwind` which is incompatible with this crate's strategy of `abort`

error: the crate `objc2` requires panic strategy `unwind` which is incompatible with this crate's strategy of `abort`

error: the crate `block2` requires panic strategy `unwind` which is incompatible with this crate's strategy of `abort`

error: aborting due to 3 previous errors

Target //backend:record failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 0.859s, Critical Path: 0.60s
INFO: 2 processes: 39 action cache hit, 2 internal.
ERROR: Build did NOT complete successfully
➜  mjolnir git:(fix-optimization-bug-egui-eframe) ✗ bazel query "somepath(//backend:record, @@rules_rust++crate+crates__objc2-0.6.4//:objc2)"
//backend:record
@crates//:ctrlc
@@rules_rust++crate+crates__ctrlc-3.5.2//:ctrlc
@@rules_rust++crate+crates__dispatch2-0.3.1//:dispatch2
@@rules_rust++crate+crates__objc2-0.6.4//:objc2
```
I'm just gonna leave panic strategy to be unwind for the 3 binaries that use `ctrlc`: record, record from one, and stream.

Here are the before and afters for each binary after I added optimizations:
| Binary | Normal Size | Optimized Size |
|-|-|-|
| Discover | 371K | 342K |
| Record From One | 2.2M | 922K |
| Record | 2.2M | 939K |
| Stream | 31M | 10M |

owens3364 (approved):

PR #41, opened 2026-04-17
PTP Synced Camera Capture!

owens3364:
Ok so hella code changes, lots of writing binary from obscure documentation and shit

Good news is, PTP synced acquisition worked! Got about 10fps from both cameras at 4k but each frame matched up to the other frame's index within about 2 microseconds which is pretty dope

Bad news is, my code is quite disgusting. It's not buggy, it's just kinda disorganized. We need to do some large refactors, just haven't gotten around to it yet.

I'm happy to clean up anything particularly egregious in this PR and then make and complete tickets for smaller refactors

BUT IT WORKS AND WE HAVE TIME SYNC SO HELL YEAH


owens3364:
Oh, I also committed the actual GVCP features and values from each cam to the repo so we can see the actual settings and their options for the camera instead of just AI hallucinating or docs guessing or being vague.

Don't be fooled by the diff it's mostly those .txt files!

backend/src/camera/aravis_utils.rs
```diff
@@ -16,14 +24,86 @@ pub fn create_camera(camera_id: &str) -> Result<Camera, String> {
...
+        }
+        return from_glib_none(raw_res);
+    }
+}
+
+fn unsafe_read_camera_boolean(camera: &Camera, node_name: &str) -> bool {
+    unsafe {
+        let mut error: *mut glib::ffi::GError = std::ptr::null_mut();
+        let camera_ptr: *mut aravis_sys::ArvCamera = camera.to_glib_none().0;
+        let feature_c_str = std::ffi::CString::new(node_name).unwrap();
```
luke-venk: nit: here you say `std::ffi::Cstring` but you import that so can just say `CString`

backend/src/camera/aravis_utils.rs
```diff
@@ -16,14 +24,86 @@ pub fn create_camera(camera_id: &str) -> Result<Camera, String> {
...
+            panic!(
+                "Error calling arv_camera_get_boolean for node: {}",
+                node_name
+            );
+        }
+        raw_res != 0
+    }
+}
+
+pub struct PtpConfig {
```
luke-venk, line 62: Should probably live in `backend/src/camera/config.rs`, right?

backend/src/camera/aravis_utils.rs
```diff
@@ -16,14 +24,86 @@ pub fn create_camera(camera_id: &str) -> Result<Camera, String> {
...
+    pub lock_barrier: CancelableBarrier,
+}
+
 /// Loads our information from our custom camera configuration type
 /// into Aravis camera.
-pub fn configure_camera(camera: &Camera, config: &CameraIngestConfig) {
+pub fn configure_camera(
+    camera: &Camera,
+    config: &CameraIngestConfig,
+    maybe_shutdown: Option<Arc<AtomicBool>>,
```
luke-venk, line 74: Is this an actual naming convention to have `maybe_` prefix `Option<T>` variables?

backend/src/camera/aravis_utils.rs
```diff
@@ -45,31 +125,155 @@ pub fn configure_camera(camera: &Camera, config: &CameraIngestConfig) {
...
+            if ptp_servo_status == target_status {
+                consecutive_successes += 1;
+            } else {
+                consecutive_successes = 0;
+            }
+            attempts += 1;
+            sleep(Duration::from_millis(500));
+        }
+        if consecutive_successes < 10 {
+            panic!("Camera {} failed to establish PtP lock", config.camera_id);
```
luke-venk: nit: `PTP` instead of `PtP`

backend/src/camera/cancelable_barrier.rs
```diff
@@ -0,0 +1,73 @@
+use std::sync::{Arc, Condvar, Mutex};
```
luke-venk, line 1: Could you please add a multi-line comment summarizing why this is needed for PTP config? Wasn't super intuitive to me at first glance

MODULE.bazel
```diff
@@ -59,6 +59,11 @@ crate.spec(
     package = "aravis",
```
luke-venk, line 59: Need to resolve merge conflicts

backend/src/camera/aravis_utils.rs
```diff
@@ -45,31 +125,155 @@ pub fn configure_camera(camera: &Camera, config: &CameraIngestConfig) {
...
+                .expect("Failed to make camera ptp slave");
+            let ptp_is_slave = unsafe_read_camera_boolean(camera, "PtpSlaveOnly");
+            if !ptp_is_slave {
+                panic!("Failed to make camera {} a PTP slave", config.camera_id);
+            } else {
+                println!("Made camera {} a PTP slave", config.camera_id);
+            }
+            let mut success = false;
+            let mut attempts = 0;
+            while !success && attempts < 300 {
```
luke-venk, line 157: This while statement, along with the ones on lines 183 and 205 could probably be extracted into a reusable helper function.

backend/src/camera/record/capture.rs
```diff
@@ -1,14 +1,120 @@
-use super::writer::{Frame, Metadata, ensure_dir, sanitize_path_name};
-use crate::camera::CameraIngestConfig;
+use super::writer::{ensure_dir, sanitize_path_name, Frame, Metadata};
```
luke-venk, line 1: nit: format imports

backend/src/camera/record/capture.rs
```diff
@@ -1,14 +1,120 @@
...
 use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{Arc};
 use std::time::{Duration, Instant};
+use crate::camera::{BarrierResult, CancelableBarrier};
+
+// Need both cams to agree on the next timestamp to start recording at
+// They're on separate threads and this code architecture doesn't make it easy for them to talk to each other
+// So we use math to agree
+fn compute_first_frame_ptp_ns(current_ptp_ns: u64) -> u64 {
+    let interval_ns: u64 = 10_000_000_000; // 10 seconds in ns
```
luke-venk, line 22: How were 10 and 3 seconds specifically chosen?

backend/src/camera/record/capture.rs
```diff
@@ -1,14 +1,120 @@
...
+    group_key: u32,
+    group_mask: u32,
+) {
+    println!("Sending action command!");
+    // GigEV action command packet: 56 bytes total
+    // Ref: GigE Vision spec section on Action Commands
+    let mut packet = [0u8; 28];
+
+    // GVCP header
+    packet[0] = 0x42; // required first byte
```
luke-venk, line 73: Where do all these numbers come from? I'm not doubting, just curious

backend/src/camera/cli_record.rs
```diff
@@ -81,4 +81,7 @@ pub struct RecordWithBothCamerasArgs {
     // Common record args.
     #[command(flatten)]
     pub common_args: CommonRecordArgs,
+
+    #[arg(long = "interface", required = true)]
+    pub interface: String,
```
luke-venk, line 86: What would this actually be in practice if both cameras have the same interface? Could you put an example usage in the `README.cameras.md`?

backend/src/bin/record_from_both_cameras.rs
```diff
@@ -74,12 +82,46 @@ pub fn main() {
...
+    let configuration_barrier_1 = configuration_barrier.clone();
+    let configuration_barrier_2 = configuration_barrier.clone();
+    let acquisition_barrier = CancelableBarrier::new(2);
+    let acquisition_barrier_1 = acquisition_barrier.clone();
+    let acquisition_barrier_2 = acquisition_barrier.clone();
+
     ctrlc::set_handler(move || {
         println!("\nShutdown signal received, stopping recording...");
         shutdown.store(true, Ordering::SeqCst);
+        ptp_enable_barrier.cancel();
```
luke-venk, line 107: Why is so much care for a cancelable barrier needed? Like why can't ctrlc just work normally on this?

backend/src/bin/record_from_both_cameras.rs
```diff
@@ -74,12 +82,46 @@ pub fn main() {
...
+        ptp_configure_barrier.cancel();
+        ptp_lock_barrier.cancel();
+        configuration_barrier.cancel();
+        acquisition_barrier.cancel();
     })
     .expect("Error setting Ctrl+C handler.");
 
+    let ptp_config_1 = PtpConfig {
+        is_slave: false,
```
luke-venk, line 116: It makes sense that we randomly choose 1 to be master, in practice how would a master really be chosen? If like one of our cameras was a better version and had a better clock?

backend/src/bin/record_from_both_cameras.rs
```diff
@@ -74,12 +82,46 @@ pub fn main() {
     let shutdown_clone1 = Arc::clone(&shutdown);
     let shutdown_clone2 = Arc::clone(&shutdown);
 
+    let ptp_enable_barrier = CancelableBarrier::new(2);
```
luke-venk, line 84: Can you summarize the purpose of each? I might be a bot but I'm still struggling with understanding this. I could just Claude it later but maybe better if you comment

luke-venk (changes requested):

backend/src/camera/aravis_utils.rs
owens3364: I agree that there is likely a better location for this but will save that for my refactor PR :)

backend/src/camera/aravis_utils.rs
owens3364: It's *my* naming convention 🤓. I do think it's reasonable though because otherwise it gets confusing. Suppose it were named `shutdown` instead of `maybe_shutdown`...

When I go to do 

`if let Some(thing) = shutdown { ... }`

what should I call "thing"? I can't call it shutdown, I guess I could call it `should_shutdown`... it gets harder with non-boolean variables though. Suppose it were `Option<SomeConfig>`. Then would I do

`if let Some(config_unwrapped) = config { ... }`

Now I have to work with `config_unwrapped` and it would be nicer to just call the actual config `config`...

Anyways, I am open to other ideas, but it is hard to name things well :)

backend/src/camera/aravis_utils.rs
owens3364: Let's leave that for the refactor too... you're not wrong but I'm not sure how I want to do it yet, and if I made a resuable function now it would involve closures.

backend/src/camera/record/capture.rs
owens3364: Ballpark :)

10 seconds is far enough into the future that the threads should all be ready by that time
3 seconds is enough margin for all threads to guaranteed agree on which 10 second interval to select, since the threads will definitely be within 3 seconds of each other on their execution path

No incredible science or math here lol but the figures work.

Post-refactor, may decide on a more reliable way to do this, although this does have the advantage of not requiring threads to communicate with each other.

backend/src/camera/record/capture.rs
owens3364: They come from the gigevision 2.0 docs, but they are not public, so I can't provide docs links which sucks. Technically we aren't supposed to commit that doc file to this repo either. I am not strongly opposed to putting it in this repo tho. I'll post or send the PDFs/docs whereever you like!

backend/src/bin/record_from_both_cameras.rs
owens3364: See the doc comment on CancelableBarrier but yeah ctrlc does not work with barriers

This is a good doc describing the origins (or near-origins) of barriers https://www.open-mpi.org/doc/v3.0/man3/MPI_Barrier.3.php

backend/src/camera/cli_record.rs
owens3364: The output of `ifconfig` for me currently includes `lo0`, `gif0`, `stf0`, `anpi1`, `anpi0`, `anpi2`, `en0`, `en1`, `en2`, `en3`, `en4`, `en5`, `en6`, `bridge0`, `ap1`, `awdl0`, `llw0`, `utun0`, `utun1`, `utun2`, `utun3`. All of those would be considered legitimate arguments here. They aren't all viable interfaces though. I'll comment :)

backend/src/bin/record_from_both_cameras.rs
owens3364: Yes, there is a negotiation method, but the guide that I linked set it up this way, and our setup doesn't have any significant differences between the two cameras (cables same type and length, cameras same firmware version and model).

MODULE.bazel
luke-venk: idk why it's not letting me put this comment on `third_party` files, but...

I've never done the work for submodules before, but do you not need a `.gitmodules` file to specify URL? And would you need updates to `third_party/BUILD.bazel` file like in trel3?

luke-venk (approved):
Approved

luke-venk (changes requested):
comment

luke-venk (approved):

PR #44, opened 2026-04-20
Vendor OpenCV

owens3364:
We have OpenCV in rust now

I did have to add LLVM toolchain but y'all will live


MODULE.bazel
```diff
@@ -17,6 +17,59 @@ module(
 )
 
 # External dependencies.
```
luke-venk, line 19: Not my forte at all, but can you please check my understanding?

LLVM: compiler infrastructure that decouples translating programming languages (frontend) to machine code regardless of target platform (backend). Why do we need this in our project?

Bindgen: Uses foreign function interface to use C++ headers to convert to Rust code

Foreign CC: How you get Cmake to build OpenCV (specifically with Clang runtime)

BUILD.bazel
```diff
@@ -1,4 +1,70 @@
+load("@bazel_skylib//rules:native_binary.bzl", "native_binary")
```
luke-venk, line 1: Again not my forte, but can you help me understand?

Skylib is just like the STL for Bazel?

You are cross compiling for musl, which is a C library for Linux? This is useful for static linking, and then we can cross compile for other targets (like Mac)?

And then you configure C version and standard

.bazelrc
```diff
@@ -10,3 +10,6 @@ build --action_env=PATH
 # See https://bazel.build/reference/be/general#config_setting_examples
 build:release -c opt
 build:release --@rules_rust//rust/settings:lto=thin
+
+build:macos --@toolchains_llvm//toolchain/config:libunwind=False
```
luke-venk, line 14: This is for mac target LLVM toolchain to not use libunwind? Why?

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "third_party/opencv"]
```
luke-venk, line 1: This is the same way you would add git submodules right? Just to be clear, vendoring is like git submodules but instead of pointing to a separate repo, it's like baking the code directly into the repo, right?

Is the reason we're using the submodule syntax that we point to some external repo to get OpenCV source code, and then Bazel compiles it locally? Is that still considered vendoring?

third_party/BUILD.bazel
```diff
@@ -0,0 +1,83 @@
+load("@rules_foreign_cc//foreign_cc:defs.bzl", "cmake")
```
luke-venk, line 1: Again, please help me make sure I understand, here we basically use Cmake to to build all the OpenCV shit we need, specified by `cache_entries`?

third_party/BUILD.bazel
```diff
@@ -0,0 +1,83 @@
...
+        "BUILD_opencv_calib3d": "ON",
+        "BUILD_opencv_features2d": "ON",
+        "BUILD_opencv_flann": "ON",
+        "BUILD_opencv_dnn": "ON",
+        # Disable everything else
+        "BUILD_opencv_highgui": "OFF",
+        "BUILD_opencv_videoio": "OFF",
+        "BUILD_opencv_video": "OFF",
+        "BUILD_opencv_ml": "OFF",
+        "BUILD_opencv_objdetect": "OFF",
```
luke-venk, line 65: Just curious, we don't need object detection stuff?

MODULE.bazel
owens3364: Mostly right.
LLVM - we need it because otherwise
1. It is really hard to suppose that *my* build of opencv will work on anyone else's machine, since Bazel defaults to using the system compiler driver and compiler/linker/archiver/etc binaries, which vary wildly machine-to-machine or may not be present at all. Using LLVM toolchain skips all of that. If OpenCV builds with OpenCV but is linking to stuff specifically on my system or using options that only work with my compiler setup, that's not nearly as reliable as if it is being compiled with a standard-across-systems LLVM toolchain
2. Rules rust bindgen needs a toolchain to be specified for use. LLVM gives us bazel targets (clang, libclang) to construct a toolchain with. Without LLVM, we would have to detect the host platform and find what is available and give specific paths to those tools, and that won't work well across-machines.

Bindgen - yes on the C++ headers, but it builds a C/C++ library file with the provided llvm tooclhain with extern c symbols added, and then generates a rust file matching the header file, and generates the rust code that is linked with the extern c symbols in the compiled c/cpp library. So the library itself is not converted, it is just modified to be linkable to a rust interface generated from the library header(s).

BUILD.bazel
owens3364: It's kind of like *a* standard library, bazel functions are kind of scattered (platforms is a separate library, the whole aspect ecosystem is fantastic but separate, rules_cc is kind of a basic essential that used to be built in but is now separate, etc). Bazel tries to keep itself pretty pure, which has tradeoffs but is a reasonable decision IMO.

Am supporting compilation for linux musl and macos yes. MUSL is a way to build static binaries for linux. Macos doesn't have quite the same thing, but the mac ecosystem is highly standardized, so it's a bit easier to not build static things and link to the system libraries on macos at runtime.

Reason for this is kind of the same as for llvm - guaranteed to work everywhere, not just on my machine. This makes it work for ppl on WSL

.bazelrc
owens3364: When we were making libzmq work in TREL, libunwind would cause build failures on mac, and after several hours of debugging, that disable line did the trick. I haven't tried it without that line, but I'd like to keep it in unless is causes problems down the line. I don't expect it to cause issues - vendoring aravis has worked well so far with that line included.

.gitmodules
owens3364: `.gitmodules` is a result of using `git submodule add`. It's not something anyone manually edits, it's maintained by `git`.

"Vendor" means "pull the code into my repo and build it myself"

git submodules is the mechanism we use to do that. It's probably the best way to. Technically, you could literally clone a repo to vendor, destroy the `.git` folder and all its VCS metadata, and then copy it into your main repo and call that vendoring.

We vendor so we can have good control over how the external libraries are built. We will need to squeeze every bit of performance out of OpenCV, and shrinking code size and symbols and debug information helps too. Also, we need bazel targets to point at and link and depend on, and many external libraries do not build cleanly with bazel, and/or need more customization.

Does that answer your questions?

third_party/BUILD.bazel
owens3364: Yes, `cache_entries` are values propagated to cmake as build settings. OpenCV, unlike many of the other libraries we depend on, is extremely well-organized and well-maintained, so using cmake is a clean (but slow) way to build it in our case, and it exposes convenient ways to customize the build (`cache_entries`). Other vendored libraries are built directly with `rules_cc`, which you'll see in my next PR.

third_party/BUILD.bazel
owens3364: We *might*, not sure yet, but it is pretty trivial to turn on what we need if we find out we don't have it :)

luke-venk (approved):
Thank you for all the explanations!

PR #45, opened 2026-04-20
Camera ingest testing

grao23:
camera ingest, currently it writes to disk and sends it to camera_ingest.rs file.

### DIFF
--- a/backend/src/camera/aravis_utils.rs
+++ b/backend/src/camera/aravis_utils.rs
@@ -5,7 +5,7 @@ use aravis::{Aravis, Buffer, Camera, Stream};
 /// discovery and recording tools.
 use std::slice;
 
-/// Retrieves token to access global state of the Aravis library.
+// Retrieves token to access global state of the Aravis library.
 pub fn initialize_aravis() -> Aravis {
     Aravis::initialize().expect("Failed to initialize Aravis.")
 }
@@ -102,6 +102,11 @@ pub fn copy_buffer_bytes(buffer: &Buffer) -> Vec<u8> {
     let (ptr, len) = buffer.data();
 
     if ptr.is_null() || len == 0 {
+        eprintln!(
+            "Warning: Aravis returned an empty buffer (ptr null: {}, len: {}).",
+            ptr.is_null(),
+            len
+        );
         return Vec::new();
     }
 
--- a/backend/src/camera/aravis_utils.rs
+++ b/backend/src/camera/aravis_utils.rs
@@ -102,7 +102,7 @@ pub fn copy_buffer_bytes(buffer: &Buffer) -> Vec<u8> {
     let (ptr, len) = buffer.data();
 
     if ptr.is_null() || len == 0 {
-        panic!("ERROR: Aravis buffer was empty");
+        return Vec::new();
     }
 
     // `ptr` is non-null and Aravis guarantees the buffer data is valid
--- a/backend/src/camera_ingest/camera_ingest.rs
+++ b/backend/src/camera_ingest/camera_ingest.rs
@@ -1,22 +1,22 @@
-use std::thread;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::time::{Duration, Instant};
 
 use aravis::{BufferStatus, CameraExt, StreamExt};
-use crossbeam::channel::{Receiver, Sender, bounded};
+use crossbeam::channel::{Receiver, RecvTimeoutError, Sender};
 
+use crate::camera::CameraIngestConfig;
 use crate::camera::aravis_utils::{
     configure_camera, create_camera, create_stream_and_allocate_buffers, initialize_aravis,
 };
-use crate::camera::{CameraIngestConfig, Resolution};
 use crate::camera::record::writer::Frame as RecordedFrame;
 use crate::camera_ingest::camera_ingest_helpers::{
     buffer_to_frame, reached_duration_limit, reached_frame_limit, recorded_frame_to_frame,
 };
-use crate::pipeline::Pipeline;
-use crate::schemas::{CameraId, Context, Frame};
+use crate::schemas::Frame as PipelineFrame;
 
 // Ingests frames from the cameras using the GigEVision API and sends them into the pipeline.
-pub fn ingest_frames(tx: Sender<Frame>, config: CameraIngestConfig) {
+pub fn ingest_frames(tx: Sender<PipelineFrame>, config: CameraIngestConfig) {
     config
         .validate()
         .unwrap_or_else(|err| panic!("invalid camera ingest config: {err}"));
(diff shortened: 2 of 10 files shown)


backend/src/camera/aravis_utils.rs
```diff
@@ -102,7 +102,7 @@ pub fn copy_buffer_bytes(buffer: &Buffer) -> Vec<u8> {
     let (ptr, len) = buffer.data();
 
     if ptr.is_null() || len == 0 {
-        panic!("ERROR: Aravis buffer was empty");
+        return Vec::new();
```
luke-venk, line 110: Wouldn't this silently fail if we get an empty buffer?

backend/src/camera_ingest/camera_ingest_helpers.rs
```diff
@@ -0,0 +1,145 @@
+use std::time::{Duration, Instant};
+
```
luke-venk, line 2: nit: get rid of all these empty lines

backend/src/camera_ingest/camera_ingest.rs
```diff
@@ -1,12 +1,285 @@
...
+use std::time::{Duration, Instant};
+
+use aravis::{BufferStatus, CameraExt, StreamExt};
+use crossbeam::channel::{Receiver, Sender, bounded};
+
+use crate::camera::aravis_utils::{
+    configure_camera, create_camera, create_stream_and_allocate_buffers, initialize_aravis,
+};
+use crate::camera::{CameraIngestConfig, Resolution};
+use crate::camera::record::writer::Frame as RecordedFrame;
```
luke-venk, line 12: Good looking out for this name collision that is my fault. I think you should just rename this `Frame` to `RecordedFrame` and likewise update name of `Frame` for pipeline to `PipelineFrame` or something

backend/src/camera_ingest/camera_ingest.rs
```diff
@@ -1,12 +1,285 @@
...
+    capacity_per_channel: usize,
+) -> (thread::JoinHandle<()>, Pipeline) {
+    start_camera_pipeline(
+        camera_id,
+        default_ingest_config(camera_id),
+        capacity_per_channel,
+    )
+}
+
+fn default_ingest_config(camera_id: CameraId) -> CameraIngestConfig {
```
luke-venk: This kind of function should exist wherever `CameraIngestConfig` is defined (in `camera/config.rs`). I think it is most clean to use the [Default](https://doc.rust-lang.org/std/default/trait.Default.html) trait to handle this.

backend/src/camera_ingest/camera_ingest.rs
```diff
@@ -1,12 +1,285 @@
...
+    }
+
+    drop(left_tx);
+    drop(right_tx);
+    left_pipeline.stop();
+    right_pipeline.stop();
+}
+
+// Starts one camera-ingest thread and one pipeline stage graph for a camera config.
+pub fn start_camera_pipeline(
```
luke-venk: A function that starts a pipeline should probably live in the file that `Pipeline` is defined (`pipeline/pipeline.rs`)

backend/src/camera_ingest/camera_ingest.rs
```diff
@@ -1,12 +1,285 @@
...
+    let ingest_handle = thread::spawn(move || {
+        ingest_frames(tx_ingest, config);
+    });
+    let pipeline = Pipeline::from_receiver(camera_id, rx_stage1, capacity_per_channel);
+
+    (ingest_handle, pipeline)
+}
+
+// Starts one default fake camera pipeline so existing startup flows still work.
+pub fn start_default_camera_pipeline(
```
luke-venk: Pretty sure function is unnecessary, why not just call `start_camera_pipeline` directly and use default Pipeline constructor?

backend/src/camera_ingest/camera_ingest.rs
```diff
@@ -1,12 +1,285 @@
...
+// Ingests frames from the cameras using the GigEVision API and sends them into the pipeline.
+pub fn ingest_frames(tx: Sender<Frame>, config: CameraIngestConfig) {
+    config
+        .validate()
+        .unwrap_or_else(|err| panic!("invalid camera ingest config: {err}"));
+
+    let start_time = Instant::now();
+    let mut frames_sent = 0usize;
+
+    if config.use_fake_interface {
```
luke-venk: I still don't really understand when `use_fake_interface` is useful. What is the use case for this?

backend/src/pipeline/pipeline.rs
```diff
@@ -14,29 +13,18 @@ pub struct Pipeline {
...
-    // create the PipelineStages, spawn each stage, and return a Pipeline.
-    // Calling this function automatically starts each of the pipeline
-    // stages.
-    pub fn new(_camera_id: CameraId, capacity_per_channel: usize) -> Self {
-        // TODO(#6): Implement Custom Queue Policy.
-
-        // Define inter-stage message channels for thread-safe message sharing.
-        let (tx_ingest, rx_stage1) = bounded::<Frame>(capacity_per_channel);
+    // Builds one pipeline stage graph around an incoming frame receiver.
+    pub fn from_receiver(
```
luke-venk: just rename this to `pub fn new...` since this is the constructor. It just so happens that a necessary argument for the constructor is the receiver, which makes sense

backend/src/camera_ingest/camera_ingest.rs
```diff
@@ -1,12 +1,285 @@
...
+        } else {
+            eprintln!("Received frame for unexpected camera {}.", source_camera_id);
+            continue;
+        };
+
+        if send_result.is_err() {
+            break;
+        }
+    }
+
```
luke-venk: You should add some `ctrlc` functionality here to spin down pipelines cleanly. Search `ctrlc::set_handler` in the main branch for examples

backend/src/camera_ingest/camera_ingest.rs
```diff
@@ -1,12 +1,285 @@
...
+pub fn run_recording_ingest(
+    frame_rx: Receiver<RecordedFrame>,
+    left_camera_id: String,
+    right_camera_id: String,
+    capacity_per_channel: usize,
+) {
+    let (left_tx, left_rx) = bounded::<Frame>(capacity_per_channel);
+    let (right_tx, right_rx) = bounded::<Frame>(capacity_per_channel);
+
+    let left_pipeline = Pipeline::from_receiver(CameraId::FieldLeft, left_rx, capacity_per_channel);
```
luke-venk: Don't think it makes sense to define pipelines in this function, shouldn't the pipelines be defined in the main files as opposed to in the `record_from_both_cameras.rs` binary?

backend/src/main.dev.rs
```diff
@@ -24,8 +24,8 @@ pub fn create_dev_app() -> Router {
 async fn main() {
     // Start the 2 pipelines (one for each camera).
     let rolling_buffer_size: usize = 10;
-    let _ = Pipeline::new(CameraId::FieldLeft, rolling_buffer_size);
-    let _ = Pipeline::new(CameraId::FieldRight, rolling_buffer_size);
+    let _ = start_default_camera_pipeline(CameraId::FieldLeft, rolling_buffer_size);
```
luke-venk: If the pipelines are also defined in `run_recording_ingest` which is called in `record_from_both_cameras.rs`, aren't there now 4 pipeline objects? I think the major refactor to figure out is how you would call `run_recording_ingest` to send frames via their tx to the pipeline tx, while also keeping pipelines in this file so you can spin them down using `ctrlc`.

luke-venk (changes requested):
Like previous PR, let's merge into `main` instead of the `dev` branch

backend/src/camera_ingest/camera_ingest.rs
grao23: I was using it for one of my test cases to check the ingest; it won't be a part of the real camera setup

luke-venk:
Replaced by PR 46: https://github.com/luke-venk/mjolnir/pull/46

PR #46, opened 2026-04-22
Real: Test Camera Ingest

### LINKED ISSUE #3: Implement Camera Ingest

grao23:
Making changes per last PR but this time with main branch

Closed #3


backend/src/camera/aravis_utils.rs
```diff
@@ -308,7 +102,12 @@ pub fn copy_buffer_bytes(buffer: &Buffer) -> Vec<u8> {
     let (ptr, len) = buffer.data();
 
     if ptr.is_null() || len == 0 {
-        panic!("ERROR: Aravis buffer was empty");
+        eprintln!(
```
luke-venk, line 332: What is the reason for switching this? I'm not saying it's wrong but yeah why?

backend/src/camera/config.rs
```diff
@@ -72,10 +84,37 @@ impl CameraIngestConfig {
         if self.num_buffers == 0 {
             return Err("num_buffers must be > 0".to_string());
         }
+        if let Some(max_frames) = self.max_frames {
```
luke-venk: Could just do 

`if self.max_frames = Some(0)`

Same for max_duration_s

backend/src/camera/config.rs
```diff
@@ -15,8 +15,11 @@ pub struct CameraIngestConfig {
     pub resolution: Resolution,
 
     // System-level config.
+    pub enable_ptp: bool,
     pub num_buffers: usize,
     pub timeout_ms: u64,
+    pub max_frames: Option<usize>,
```
luke-venk: Why would Aravis or camera_ingest care about max_frames or max_duration_s? These are only parameters for recording. Our pipeline would never want to stop working just bc of some frame or time duration.

TL;DR: CameraIngestConfig was meant to be for how Aravis handles camera specs for getting footage. So why would it care about max duration or time, which are CLI args for just recording?

backend/src/camera_ingest/camera_ingest.rs
```diff
@@ -1,12 +1,225 @@
-use crate::schemas::{Context, Frame};
+use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::time::{Duration, Instant};
```
luke-venk, line 8: nit: get rid of all these newlines

backend/src/bin/record_from_both_cameras.rs
```diff
@@ -68,17 +70,29 @@ pub fn main() {
...
 
-    // Create crossbeam channel so capture thread can send frames to
-    // write thread. Have a clone of the same sender, as well as the
-    // original sender, send to the receiver.
+    // Create crossbeam channel so capture threads can send frames to
+    // the fan-out thread.
     let (frame_tx1, frame_rx) = crossbeam::channel::bounded::<Frame>(100);
     let frame_tx2 = frame_tx1.clone();
 
+    // The fan-out thread duplicates each captured frame so one copy can
```
luke-venk: Major refactor: No ingest or pipeline stuff should exist in this binary.

We need it so that the main binaries, if in `real` mode (i.e. actually want footage from cameras instead of simulating it) call the recording code somehow and ingest that into their pipeliens

luke-venk (changes requested):

backend/src/camera/aravis_utils.rs
grao23: wanted to make sure it still ran, so took out thr panic, and said if it was an error then say that statement so we would still know

backend/src/bin/record_from_both_cameras.rs
grao23: ok refactored it to fit this

luke-venk:
Run this command to test (per Gautam)

`bazel run //backend:prod_real -- --feed-footage-dir (folder)`

luke-venk:
Closed in favor of #78

Issue #4, opened 2026-03-04
Rewrite Computer Vision in Rust


luke-venk:
## Description
While we will tune our CV in Python, our final product should have the algorithm implemented in our Rust pipeline. Once the CV is tuned, we need to urgently write it into our pipeline.

## Setup
1. Here is the the official Rust book: [The Rust Programming Language](https://doc.rust-lang.org/book/). This is super useful. I shouldn't have to convince you too hard to read this like I would other people, because I know you can handle reading documentation :). But overall I find it an easy read compared to other docs, and it explains concepts very well and has examples. I would at least read chapter 1 to handle installation, and then refer to it as needed. If you ask me questions I am happy to answer them and send you relevant pages in this docs.
2. I think it will be easiest for you if you download Windows Substitute for Linux (WSL). I'm not sure if you're familiar with it, but developing software like ours is very difficult for Windows (don't even ask Owen about this, it's a touchy subject especially since our cursed Best Buy trip), so we only support Linux and Mac. Let Gautam know if you need help, since he just did this.
3. Once you have WSL, go to our repository and run the setup script: `./scripts/setup_linux.sh`. This should take care of dependencies. If you have any issues going forward, let me know, or just look up what to install and update the setup script accordingly.

## Steps
1. Determine each of the stages of CV that we perform. Right now Owen has mentioned undistortion, intensity normalization, forward downsampled copy, mog2, and contour.
2. You will need to use the [opencv Rust crate](https://crates.io/crates/opencv) to interface with OpenCV in Rust. Update `MODULE.bazel` and `backend/BUILD.bazel` (see below section for more info) with this crate and any other crate you need.
3. Update the files found in `backend/src/computer_vision/` (see below section for more info). Each of these files contain functions that take in a Frame (a vector of bytes and a timestamp basically, see `backend/src/schemas/frame.rs`), performs some CV operation (you need to fill this in), and returns a new Frame. This is the vast majority of your work.
4. Update `backend/src/pipeline/pipeline.rs` to include all your new CV stages (see below section for more info)
5. I think it would be good to write unit tests for your code to ensure it works. I'm not 100% sure how to go about this or if this is viable, but I think a decent place to start is to write tests to confirm that, for the same input, the Rust CV returns the same thing the Python CV would have returned. We can cross this bridge once it comes to it, but just a thought, and what I asked Yash to do in #9.
[...]

PR #47, opened 2026-04-22
Implement Downsampling in Rust Computer Vision

luke-venk:
Related to #4

Sorry for the big diff. Summary of changes:
- I refactored `Resolution` to `AtlasATP124SResolution` resolution, which is why you see diffs in the `camera` directory. I also changed 4k -> Full, 1080p -> Half, and 720p -> Quarter, since technically nothing we were doing was any of 4k, 1080, or 720p, which are terms mostly for TV
- I got rid of the `schemas` directory and just moved stuff to the `pipeline` directory
- Implemented `forward_downsampled_copy`
- Added `CameraId` to `Context` so it's a unique hash
- Updated `Frame` to how we discussed and updated unit tests throughout
- Update unit test helpers to populate the `Mat`s in `Frame` as we would like
- Updated setup scripts to automatically run `git submodule update...`
- Update `third_party/BUILD.bazel` to fix linker issues
- Update `README.cameras.md`


backend/src/computer_vision/forward_downsampled_copy.rs
```diff
@@ -1,9 +1,74 @@
...
+        let data: Vec<u8> = vec![69u8; (h * w) as usize];
+        let input_frame: Frame = Frame::new(data, context);
+        
+        let output_frame = forward_downsampled_copy(input_frame);
+
+        let expected_size: usize = (DOWNSAMPLED_WIDTH_PX * DOWNSAMPLED_HEIGHT_PX) as usize;
+        assert_eq!(output_frame.data().len(), expected_size);
+        assert_eq!(output_frame.context().resolution(), Resolution::Downsampled);
+    }
+}
```
luke-venk: nit: add newline

luke-venk (commented):
smh after all these reviews where i ask for newlines i forgot one

scripts/setup_linux.sh
```diff
@@ -110,4 +110,7 @@ echo "Generating Rust project..."
 cd "$PROJECT_ROOT"
 bazel run @rules_rust//tools/rust_analyzer:gen_rust_project
 
+echo "Updating git submodules..."
+git submodule update --init --recursive
```
owens3364: Add --depth 1 so that it only fetches the minimum required recent comments instead of the entire repo history

scripts/setup_mac.sh
```diff
@@ -103,5 +103,7 @@ echo "Generating Rust project..."
 cd "$PROJECT_ROOT"
 bazel run @rules_rust//tools/rust_analyzer:gen_rust_project
 
+echo "Updating git submodules..."
+git submodule update --init --recursive
```
owens3364: Add --depth 1 so that it only fetches the minimum required recent comments instead of the entire repo history

backend/src/camera/config.rs
```diff
@@ -77,14 +77,22 @@ impl CameraIngestConfig {
...
+#[derive(Debug, Clone, Copy, ValueEnum, Eq, PartialEq, Hash, Default)]
 pub enum Resolution {
     #[value(name = "720p")]
     HD,
+
     #[value(name = "1080p")]
     FullHD,
+
     #[value(name = "4k")]
+    #[default]
```
owens3364, line 92: Honestly didn't know you could do this, neat!

backend/src/camera/config.rs
```diff
@@ -102,6 +111,7 @@ impl Resolution {
             Resolution::HD => 4,
             Resolution::FullHD => 2,
             Resolution::UHD4K => 1,
+            Resolution::Downsampled => panic!("Error: Do not use Downsampled resolution for binning!"),
```
owens3364: IMO this shouldn't be a panic

Panic is for deep, unacceptable logic issues or invalid state that we can't easily recover from

This is more forcing users into a certain decision, and IMO we might want to be able to stream a downsampled variant... I don't see a technical reason that that shouldn't be possible.

backend/src/camera/config.rs
owens3364, line 1: On further thought, if we can't use Downsampled for streaming from the camera, and it doesn't fit into a binning integer anyway, I'm not sure it should be a camera config setting at all...

backend/src/computer_vision/forward_downsampled_copy.rs
```diff
@@ -1,9 +1,74 @@
...
+            frame.data().as_ptr() as *mut _,
+            opencv::core::Mat_AUTO_STEP,
+        )
+        .expect("Error: Failed to generate input Mat from frame in forward_downsampled_copy().")
+    };
+    let mut output_mat: Mat = Mat::default();
+
+    if let Err(_) = resize(&input_mat, &mut output_mat, SIZE, 0.0, 0.0, INTER_LINEAR) {
+        eprintln!(
+            "Error: Failed to downsample frame in forward_downsampled_copy(). Returning original frame."
```
owens3364: Put the error value into this message instead of ignoring it with `_`

backend/src/computer_vision/forward_downsampled_copy.rs
```diff
@@ -1,9 +1,74 @@
...
+        let (h, w): (i32, i32) = resolution.dimensions();
+        let context: Context = Context::new(6969, resolution);
+        let data: Vec<u8> = vec![69u8; (h * w) as usize];
+        let input_frame: Frame = Frame::new(data, context);
+        
+        let output_frame = forward_downsampled_copy(input_frame);
+
+        let expected_size: usize = (DOWNSAMPLED_WIDTH_PX * DOWNSAMPLED_HEIGHT_PX) as usize;
+        assert_eq!(output_frame.data().len(), expected_size);
+        assert_eq!(output_frame.context().resolution(), Resolution::Downsampled);
```
owens3364: Can you also assert that everything in the arr is `69` ? I want to make sure the data isn't lost or mangled in this transformation.

backend/src/pipeline/frame.rs
```diff
@@ -0,0 +1,69 @@
...
+        self.timestamp
+    }
+
+    pub fn resolution(&self) -> Resolution {
+        self.resolution
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Default)]
+pub struct Frame {
```
owens3364: Can we add an `id` i32 property? Will want to map a frame back to its original high res data at some point when we know the trajectory and need to refine our position estimates.

backend/src/computer_vision/forward_downsampled_copy.rs
```diff
@@ -1,9 +1,74 @@
...
+use crate::pipeline::{Context, Frame};
+use opencv::core::{CV_8U, Mat, Size};
+use opencv::imgproc::{INTER_LINEAR, resize};
+use opencv::prelude::MatTraitConstManual;
+
+const DOWNSAMPLED_WIDTH_PX: i32 = 960;
+const DOWNSAMPLED_HEIGHT_PX: i32 = 540;
+const SIZE: Size = Size::new(DOWNSAMPLED_WIDTH_PX, DOWNSAMPLED_HEIGHT_PX);
 
 pub fn forward_downsampled_copy(frame: Frame) -> Frame {
```
owens3364, line 15: We should be careful about how this function deals with and manages memory.
1. I don't know if the Mat construction on 23 is copying bytes out of the frame, or borrowing memory without a lifetime, or taking ownership of that memory. Also, it looks like a pointer to mutable data is being used - or I am wrong about the syntax meaning, that's possible too. But the data shouldn't be modified with this, so I am curious about that.
2. If there is a more efficient way to do this, we should explore that. A better approach might be to make `Frame` natively contain a `Mat`, and then we don't have to copy bytes in and out for every pipeline stage.

backend/src/lib.rs
```diff
@@ -1,8 +1,14 @@
+#[cfg(feature = "real")]
 pub mod camera;
+#[cfg(feature = "real")]
 pub mod camera_ingest;
+// Technically it doesn't make sense to include circle infractions code in fake lib, but will break if not included. Spin off ticket?
```
owens3364: Def spinoff ticket, althought it is likely a simple enough fix that you could handle it in this PR

owens3364 (changes requested):

backend/src/camera/config.rs
luke-venk: I defined `Resolution::Downsampled` because `Frame` requires a `Context`, which has a `Resolution`. The `Context` needs `Resolution` because when the resizing requires `forward_downsampled_copy()` to have the original resolution of the image. I figured the easiest way to do this was to include the frame's resolution in the metadata (`Context`). And I figured the simplest way to do that was just reuse the `Resolution` enum, if that makes sense.

Do you think I should refactor it in a different way just to avoid panicking? I think it makes sense to fail if anyone tries to bin or to_string that resolution, since if the user tries to record with that nonsensical resolution it makes no sense.

backend/src/computer_vision/forward_downsampled_copy.rs
luke-venk: Good point. I was struggling to find a way to check that the image wasn't corrupted in this unit test, but couldn't think of a reasonable way. This test isn't super robust but I don't think it has to handle that as opposed to ensuring downsampling to lower resolutions works.

backend/src/pipeline/frame.rs
luke-venk: How would I know what frame ID should actually be? Should it be based on the order in which they come via the `recording` or `ingest` pipelines? And then through various CV pipeline stages do they keep their original ID?

backend/src/computer_vision/forward_downsampled_copy.rs
luke-venk: Here are the docs for `new_rows_cols_with_data_unsafe`
> /// * data: Pointer to the user data. Matrix constructors that take data and step parameters do not
> /// allocate matrix data. Instead, they just initialize the matrix header that points to the specified
> /// data, which means that no data is copied. This operation is very efficient and can be used to
> /// process external data using OpenCV functions. The external data is not automatically deallocated, so
> /// you should take care of it.

This seems to imply it's borrowing memory and just viewing the data.

As for the pointer to mutable data, the docstring says it doesn't allocate matrix data. Not sure why it's like this but it should be fine.

TL;DR, I think it's fine as is, but it is probably cleaner to factor. I was actually gonna suggest we just have `Frame` natively contain `Mat` instead of `data` if `Mat` is used in the majority of OpenCV functions. I'll do that now.

backend/src/lib.rs
luke-venk: I cut #52. Tbh I'd prefer to not muddy this PR up with those changes.

backend/src/camera/config.rs
owens3364: I think `Resolution` clearly pertains to Camera settings, but that will not scale well to OpenCV because, while we can only set a couple native camera resolutions (binning is integers), we can set any arbitrary resolution with OpenCV. We may wind up doing that, too, depending on throughput and performance and how undistortion actually behaves.

So, because the frame resolution is being determined by different things (camera binning vs openCV manipulation), I think it makes sense to separate their data types.

`Context` should contain a `(u32, u32)` or `(u16, u16)` or `(usize, usize)` resolution property instead of borrowing from camera resolutions.

If we wanted to be really good about resolution specificity and naming and allat we would name our current `Resolution` enum `AtlasATP124SNativeResolution` since that is where all our pixel values are actually defined from (4k means different resolutions on different cameras).

backend/src/computer_vision/forward_downsampled_copy.rs
owens3364: Ok after looking into Mat pretty extensively...
What you are doing here below is safe, but only because you are copying the result into a new buffer that the `output_mat` actually owns. Your zero-copy `input_mat` creation is good for performance, but if the underlying buffer that you create it from is freed by rust / goes out of scope, and then you do anything with the underlying buffer, you'll be accessing freed memory and it'll crash.

This isn't actually anything wrong with your code right now, but something to keep in mind bc we can't tell by looking if the Mat owns its memory or is pointing elsewhere.

Now, more significantly for this PR - why does this code create an `input_mat` at all, when there's already the `Mat` in `Frame`?

backend/src/camera/config.rs
owens3364: Edit: since `Frame` contains `Mat` and `Mat` contains resolution information, we should drop `resolution` from `Context` entirely

backend/src/pipeline/frame.rs
owens3364: Actually good point because then if frames show up out of order we'll have a weird ID scheme. `Context` is already `Hash`. Let's add the camera id camera left / camera right thing to it, and then the hash of `Context` will be a fully unique and meaningful id.

backend/src/camera/config.rs
luke-venk: Sounds good. I'll make the change for `Resolution` by renaming and refactoring a bit. Tbh the terms 720p, 1080p, and 4k are misleading since those imply 16:9 aspect ratios which are more for TVs and not what our cameras use, so I'll refactor those as well.

backend/src/pipeline/frame.rs
```diff
@@ -0,0 +1,80 @@
...
+        self.timestamp
+    }
+
+    pub fn resolution(&self) -> Resolution {
+        self.resolution
+    }
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct Frame {
```
owens3364: `Frame` should probably contain
- `raw_full_res_frame_bytes` as a `Box<[u8]>`
- `raw_full_res_frame_resolution` as a `(int, int)` where int is either u16 or u32 or i32 or usize
- `undistored_full_res_image` as a `OnceLock<Mat>`
- `undistorted_downsampled_image` as a `OnceLock<Mat>`

We'll likely add more as we populate more pipeline stages, but that's a decent getting started.

Please add some large descriptive rustdoc comments to the `Frame` type explaining that
- `Mat` vars in here should own their data as opposed to having a unsafe mutable pointer to another data buffer, like the `input_mat` above
- Any operations using `Mat` **cannot** modify the underlying data when it is owned elsewhere. Ie, if I have a `Box[u8]` and I pass a raw pointer to it to a `Mat` unsafe constructor, and then have the `Mat` mutate that data, that's real bad... noooooooo doing that.

backend/src/computer_vision/forward_downsampled_copy.rs
luke-venk: Good catch. It was written like that before `Frame` owned a `Mat`, which is why I was like generating the `Mat` from it's vector of bytes. Changing

backend/src/computer_vision/forward_downsampled_copy.rs
owens3364: I would strongly recommend validating that the data in your new `Mat` is what you expected it to be. Depending on compilation options, the array backing your output `Mat` will default to a bunch of uninitialized memory or a bunch of zeros.

While I would not expect any issues with OpenCV doing this task, checking that at least a couple elements of your resulting `Mat` are `69` proves that it isn't a default-value `Mat` and that the information was copied correctly, and given that we are doing some hacky `unsafe` mutable pointer shit, we should validate that this works correctly.

If you want to do that in a separate test, that's fine, but I think it's a valuable check.

backend/src/pipeline/frame.rs
owens3364: Not quite sure why this is resolved - maybe you forgot to push?

luke-venk: If we do it like this, where `Frame` just stores each of the `Mat` after each CV stage works on them, that totally changes our structure where `PipelineStages` accept a function that takes in frames and returns frames, right? Since instead it just updates the contents of the same `Frame`?

luke-venk: Update: will just return the same frame after modifying

backend/src/computer_vision/forward_downsampled_copy.rs
```diff
@@ -1,9 +1,68 @@
...
+        eprintln!(
+            "Error: Failed to downsample frame in forward_downsampled_copy(). Returning original frame. {err}"
+        );
+        return frame;
+    }
+
+    // Set the downsampled image to the result and return the frame.
+    // Note: I know that we are throwing away the result but idk if I should
+    // handle this now or not.
+    frame.set_downsampled_image(output_mat).unwrap();
```
owens3364: add expect

backend/src/computer_vision/mog2.rs
```diff
@@ -1,12 +1,11 @@
...
 
 pub fn mog2(frame: Frame) -> Frame {
-    // TODO: implement actual logic
+    // TODO: Currently just passes the frame through this stage untouched.
+    // Please implement the actual Mog2 logic.
 
-    let data = frame.data();
-    let new_timestamp = frame.context().timestamp() + 1;
-    Frame::new(data.to_vec(), Context::new(new_timestamp))
+    frame.clone()
```
owens3364: drop all `frame.clone()`s as returns of `fn (frame: Frame) -> Frame`-signatured functions

backend/src/pipeline/frame.rs
```diff
@@ -0,0 +1,174 @@
...
+/// while only being written once.
+#[derive(Debug, Clone)]
+pub struct Frame {
+    /// The raw bytes for the frame coming from the camera at full resolution.
+    raw_bytes_full_resolution: Box<[u8]>,
+
+    /// The (width, height) of pixels for the frame. Note that this shouldn't
+    /// be confused with (rows, cols), which is in fact the opposite but
+    /// is the convention for matrices.
+    raw_full_resolution: (i32, i32),
```
owens3364: Switch to u32

backend/src/pipeline/frame.rs
```diff
@@ -0,0 +1,174 @@
...
+        self.raw_full_resolution
+    }
+
+    pub fn undistorted_image(&self) -> Option<&Mat> {
+        self.undistorted_image.get()
+    }
+
+    pub fn set_undistorted_image(&self, mat: Mat) -> Result<(), String> {
+        self.undistorted_image
+            .set(mat)
```
owens3364, line 68: Just use `?` and the default `OnceLock.set` error type in your `Result` return type above

backend/src/pipeline/frame.rs
```diff
@@ -0,0 +1,174 @@
...
+            .map_err(|_| "Undistorted image already set.".to_string())
+    }
+
+    pub fn intensity_normalized_image(&self) -> Option<&Mat> {
```
owens3364: No more intensity normalization

backend/src/pipeline/frame.rs
```diff
@@ -0,0 +1,174 @@
...
+        self.intensity_normalized_image
+            .set(mat)
+            .map_err(|_| "Intensity normalized image already set.".to_string())
+    }
+
+    pub fn downsampled_image(&self) -> Option<&Mat> {
+        self.downsampled_image.get()
+    }
+
+    pub fn set_downsampled_image(&self, mat: Mat) -> Result<(), String> {
```
owens3364, line 76: Just use `?` and the default `OnceLock.set` error type in your `Result` return type above

backend/src/pipeline/frame.rs
```diff
@@ -0,0 +1,174 @@
...
+        self.downsampled_image
+            .set(mat)
+            .map_err(|_| "Downsampled normalized image already set.".to_string())
+    }
+
+    pub fn context(&self) -> Context {
```
owens3364: Nit: Use `&Context`

backend/src/pipeline/frame.rs
```diff
@@ -0,0 +1,174 @@
...
+pub enum CameraId {
+    FieldLeft,
+    FieldRight,
+}
+
+/// Our metadata for each frame.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Context {
+    camera_id: CameraId,
+    timestamp: u64,
```
owens3364: Be specific on what timestamp this is - I am guessing this is image capture timestamp reported by camera but we should be specific

backend/src/pipeline/frame.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::camera::AtlasATP124SResolution;
+    use crate::pipeline::test_utils::{ComputerVisionStage, generate_frame};
+
+    #[test]
```
owens3364: `#[test]` should be replaced by `#[rstest]`

owens3364 (changes requested):

backend/src/pipeline/frame.rs
luke-venk: Can't use `?` because the error type of `set(mat)` is the `Mat` itself, but our error type is a `String`

owens3364 (approved):

PR #55, opened 2026-04-24
Circle Infractions Integration with Server

EloghosaE:


backend/src/server/server.rs
```diff
@@ -199,4 +232,5 @@ mod tests {
         let response = app.oneshot(request).await.unwrap();
         assert_eq!(response.status(), StatusCode::BAD_REQUEST);
     }
+
```
luke-venk: nit: rm empty line

backend/src/server/server.rs
```diff
@@ -54,18 +54,46 @@ async fn get_throw_type(State(state): State<AppState>) -> Json<GetThrowTypeRespo
...
+    let history = state.get_infraction_history().await;
+
+    // Use the most recent infraction only (simpler + less confusing)
+    let recent_circle = history
+        .last()
+        .map(|t| now_ms.saturating_sub(*t) <= 10_000) // 10 seconds
+        .unwrap_or(false);
+
+    if recent_circle {
+        result.infractions.push(Infraction {
```
luke-venk, line 103: Correct me if I'm wrong, but at the moment `get_random_infractions()` in `simulate_throw.rs` has a 30% chance of pushing a circle infraction with a random confidence value.

If that 30% occurs, and on the actual circle infractions system there is NO infraction, then I'm pretty sure this would lead to the simulated event still saying there is an infraction, which is not what we want.

Can you make it such that if `recent_circle` is true, it will add the circle infraction, but if there is no infraction, you make sure that the result deletes any circle infractions if they exist?

backend/src/server/server.rs
```diff
@@ -54,18 +54,46 @@ async fn get_throw_type(State(state): State<AppState>) -> Json<GetThrowTypeRespo
...
+    };
+
+    // 2) If we saw a recent circle infraction signal, attach it to this throw
+    let now_ms = chrono::Utc::now().timestamp_millis() as u64;
+    let history = state.get_infraction_history().await;
+
+    // Use the most recent infraction only (simpler + less confusing)
+    let recent_circle = history
+        .last()
+        .map(|t| now_ms.saturating_sub(*t) <= 10_000) // 10 seconds
```
luke-venk: @owens3364 I'd like to hear what you think, but it sounds like hardcoding 10 seconds is not what we want to attach an infraction to a throw, right? I thought we mentioned something like we'd have the user specify throw start and end, and if the circle infraction timestamp is within those start and end timestamps, that's how we know it's an infraction

backend/src/server/server.rs
```diff
@@ -54,18 +54,46 @@ async fn get_throw_type(State(state): State<AppState>) -> Json<GetThrowTypeRespo
...
+    if recent_circle {
+        result.infractions.push(Infraction {
+            infraction_type: InfractionType::Circle,
+            confidence: 1.0,
+        });
     }
+
+    Json(result)
 }
+async fn get_circle_status(State(state): State<AppState>) -> Json<serde_json::Value> {
```
luke-venk, line 112: nit: add 1 newline above

backend/src/server/server.rs
```diff
@@ -54,18 +54,46 @@ async fn get_throw_type(State(state): State<AppState>) -> Json<GetThrowTypeRespo
...
+    let stale = state.is_circle_infraction_system_stale().await;
+    let history = state.get_infraction_history().await;
+    let last_infraction_ms = history.last().copied();
 
+    Json(json!({
+        "stale": stale,
+        "last_infraction_ms": last_infraction_ms,
+        "history_len": history.len(),
+    }))
+}
```
luke-venk, line 122: nit: add 1 newline below

luke-venk (changes requested):

backend/src/server/server.rs
owens3364: Your first question is correct - yeah, no hardcoding 10s.
Process will be
1. User presses 'analyze throw'
2. We go over our 60s ringbuffer and find the confident, most recent, trajectory that we have visualized
3. We do the analysis, distance estimation, etc over that trajectory
4. We correlate the infraction history with the timestamps and look for circle infractions between [throw start - 2 seconds, throw end]
5. If there was an infraction within that duration, we note it

Now, we currently have no way to relate camera PTP times to local times, and camera frames are in PTP time and circle infractions are in local time. After looking into this...
- I don't want users to specify the throw start and end *just* to solve this problem... we can consider that down the road for more significant issues, but IMO not worth it for this.
- Getting computers to run PTP services and sync their default os system time syscall to the ptp clock will not be reliable across systems, is not easily bundlable into a static binary, and almost definitely won't work over WSL...
- So, we should include a startup period where we correlate the computer time with the camera ptp times and build a *somewhat* reliable correlation (<100ms), and then use that to relate PTP timestamps to circle infraction system timestamps
- I will make stubs for that and you @EloghosaE can integrate their usage into this PR

backend/src/camera/record/capture.rs
```diff
@@ -1,3 +1,5 @@
+use crate::time::ptp_offset;
+
```
owens3364: Nit: lose the newline

backend/src/camera/record/capture.rs
```diff
@@ -252,6 +266,9 @@ pub fn run_capture_thread(
     let start_time: Instant = Instant::now();
     let mut frames_saved: usize = 0usize;
 
+    // NEW: periodic offset update (once/sec)
```
owens3364: Don't need 'NEW' comments please lose them

backend/src/camera/record/capture.rs
```diff
@@ -279,10 +305,6 @@ pub fn run_capture_thread(
         // Stop streaming if a maximum duration was configured and the camera has recorded
         // for that amount of time. Note that the max duration needs to account for both
         // time spent throwing away frames and time spent recording still frames for Mog2.
-        // Since the time duration for recording for Mog2 (mog2_duration_s) is an approximation
```
owens3364, line 282: Why is this deleted?

backend/src/camera/record/capture.rs
```diff
@@ -291,8 +313,6 @@ pub fn run_capture_thread(
             }
         }
 
-        // If the elapsed time has not passed the throwaway duration, print out to the user
```
owens3364, line 294: Why is this deleted?

backend/src/camera/record/capture.rs
```diff
@@ -303,15 +323,12 @@ pub fn run_capture_thread(
...
                 countdown_timer = Instant::now();
             }
-            // Also be sure to drain the buffer during this time.
             if let Some(buffer) = stream.timeout_pop_buffer(0) {
                 stream.push_buffer(buffer);
             }
             continue;
         }
 
-        // Load camera buffer.
```
owens3364, line 313: Why is this deleted?

backend/src/camera/record/capture.rs
```diff
@@ -320,11 +337,6 @@ pub fn run_capture_thread(
                 buffer
             }
             None => {
-                // If the buffer hasn't yet loaded, skip.
```
owens3364, line 323: Why is this deleted?

backend/src/camera/record/capture.rs
```diff
@@ -335,17 +347,10 @@ pub fn run_capture_thread(
             }
         };
 
-        // If loading the buffer worked, copy the frame buffer into raw bytes.
         match buffer.status() {
             BufferStatus::Success => {
                 let elapsed_since_start = start_time.elapsed();
 
-                // If we still haven't saved the number of frames required for Mog2 to
```
owens3364, line 343: Why is this deleted?

backend/src/circle_infractions_ingest/circle_infractions_thread.rs
```diff
@@ -56,118 +64,82 @@ fn find_arduino_port() -> String {
     }
 }
 
-/// Finds the arduino on a serial port and begins reading in data
-/// Returns a channel of CircleInfractionDetectionState updates, which will be
```
owens3364, line 60: Why are you deleting doc comments

owens3364 (changes requested):

PR #59, opened 2026-04-25
Implement contours for CV

rx9933:
Adding code for identifying contours for shotput detection. Note that when I run the unit tests, I have errors in forward_donwample_copy.rs regardless of whether it is the blank file (original on main) or the code Luke has.


analysis/opencv_approach/tuned_cv.py
```diff
@@ -0,0 +1,686 @@
+"""
```
luke-venk, line 1: Think this file should be handled in PR #29. Can you please resolve it there?

luke-venk (changes requested):
I didn't look more in detail yet, just a small comment first to close out the other PR

luke-venk:
Closed in favor of #83

PR #60, opened 2026-04-26
Vendor Aravis

### LINKED ISSUE #43: Vendor Aravis and OpenCV
## Description
Vendoring means to copy a dependency's source code directly into our own repo, instead of fetching it from online at build time. The distinction between what libraries need vendoring vs. ones that don't are whether they are pure Rust or not. For example, we need to vendor Aravis and OpenCV because they are Rust bindings to stuff written in C/C++; they have their own non-Rust build systems (Cmake, etc.) and system dependencies that Bazel won't know how to handle. 

If we vendor these, this will let us control exactly which version/configuration gets built, and we can ensure hermetic builds.

## Work Needed
Probably some super complicated shit

owens3364:
yea type shit
Closes #43


third_party/generation.bzl
```diff
@@ -0,0 +1,88 @@
+"""Helpers for generating glib headers"""
+
+def generate_visibility_header(name, namespace, version, out):
+    """
+    Generates the visibility headers for glib/ libraries
+
+    Args:
+        name: fuck
```
luke-venk, line 8: nice

.bazelignore
```diff
@@ -0,0 +1,3 @@
+third_party/pcre2
```
luke-venk, line 1: PCRE2 is regex? Why do we need this?

BUILD.bazel
```diff
@@ -9,6 +9,30 @@ config_setting(
     values = {"compilation_mode": "opt"},
 )
 
+config_setting(
```
luke-venk, line 12: Are these the lines of code that make it so we can build on the most common types of computer architectures?

MODULE.bazel
```diff
@@ -107,20 +107,10 @@ crate = use_extension(
 )
 
 # Add external Rust dependencies here.
-crate.spec(
-    default_features = False,
-    package = "aravis",
```
luke-venk, line 112: You can remove this because now in `backend/BUILD.bazel` you specified this as a dep?
```
"//third_party:aravis",
```

So is the idea that if you are vendoring using submodules, you don't need to specify that library in `MODULE.bazel`?

luke-venk (approved):
Just to clarify, the summary of this long ass PR is this right?
- All these auto generated files are just required for building Aravis with it's build tool Meson right?
- You vendor a bunch of libraries needed for all this using submodules?
- You update our Bazel files to use these vendored submodules

BUILD.bazel
owens3364: Yes- our Mac users use arm but our WSL ppl are on x86. It is nice to include arm Linux support either for future CICD, future Linux on ARM machines, or for validating that you can build a binary for Linux despite being on an arm machine. Since some of the Cpp libraries need different args depending on target OS and architecture, we need to differentiate the configurations that way and apply correct settings using select statements.

MODULE.bazel
owens3364: Yes

The crate.spec in MODULE.bazel gives the rules rust crate universe the tool to make its own Cargo.toml / equivalent, resolve all the cargo deps, and then generate bazel build files for them, and then build everything.

if we let cargo do that for aravis and glib and allat, it will generate cargo deps trees for those things and rely on system libraries. So instead, we vendor and supply our own set of build args.

I actually vendored the aravis and glib rust crates too, but you’ll notice that I excluded their build.rs scripts so that I could hack the build together myself with Bazel.

third_party/generation.bzl
owens3364: Lmao I forgot abt this I was just trying to shut bazel up

I’ll write better docs in a future PR….

.bazelignore
owens3364: You are correct. It is a C++ dependency of glib, so we need to pull it in, build it, and give it to glib.

it is ignored because it contains its own Bazel files so Bazel sees that and gets confused. Have to tell Bazel to ignore the bazel files in that specific directory tree.

PR #61, opened 2026-04-26
Split fake infractions and fake cams

### LINKED ISSUE #58: Decouple Testing Real Inputs Between Circle Infractions and Cameras
## Description
Currently, we have the `real` build configuration that means we don't simulate inputs. But at the moment that requires us to both have real footage from cameras and real sensors for circle infraction data.

## Work Needed
Refactor build so that can test those systems separately. Want:
1. Everything simulated
2. Real camera, simulated circle infraction
3. Real circle infraction, simulated CV
4. Real both

owens3364:
Closes #58


backend/src/circle_infractions_ingest/circle_infractions_thread.rs
```diff
@@ -1,9 +1,9 @@
-#[cfg(feature = "real")]
+#[cfg(feature = "real_infractions")]
```
luke-venk: nit: do you think it would look nicer as `real_circle` as opposed to infractions, which technically someone could argue means either a circle infraction OR a sector violation?

plus maybe looks nicer to have `real_circle` since we also use `real_cameras`, and they both apply to their "sensors" that way

backend/BUILD.bazel
```diff
@@ -50,7 +92,49 @@ rust_library(
 )
 
 rust_library(
-    name = "backend_lib_fake",
+    name = "backend_lib_real_infractions",
```
luke-venk: same as previous comment, maybe "circle" instead of "infractions"?

backend/BUILD.bazel
```diff
@@ -50,7 +92,49 @@ rust_library(
...
+        ],
+    ),
+    crate_features = ["real_infractions"],
+    crate_name = "backend_lib",
+    rustc_flags = select({
+        "//:release": ["-Cstrip=debuginfo"],
+        "//conditions:default": [],
+    }),
+    deps = [
+        "@crates//:aravis",
```
luke-venk: I know this is how it was set up already, but I feel like the libraries with simulated throw data woudln't need `aravis` right? Maybe i just missed that a long time ago, in which case we can just ignore

luke-venk (changes requested):

luke-venk (changes requested):
Rationale for suggestion of naming to `circle` instead of `infractions`

In [schemas.ts](https://github.com/luke-venk/mjolnir/blob/main/frontend/lib/schemas.ts#L3-L7), we defined infraction as either sector violations or circle:
```
export enum Infraction {
  LEFT_SECTOR = "left_sector",
  RIGHT_SECTOR = "right_sector",
  CIRCLE = "circle",
}
```

Same for [infractions.rs](https://github.com/luke-venk/mjolnir/blob/main/backend/src/throws/infractions.rs#L5-L9) (I did copy you after all lol):
```
pub enum InfractionType {
    LeftSector,
    RightSector,
    Circle,
}
```

This also isn't the end of the world, I just think it's a bit clearer. Also there are merge conflicts in BUILD.bazel

luke-venk (approved):

PR #62, opened 2026-04-26
update website

### LINKED ISSUE #33: Update Frontend
## Description
- [x] Update the simulation logic to reflect changes made in this PR: See relevant [PR comment](https://github.com/luke-venk/mjolnir/pull/16#discussion_r3068571162).
- [x] Currently frontend has 3 images. Need to downsize to 2 images since we have 2 cameras.
- [x] Currently the infractions have associated confidences with them from 0.0 - 1.0. I think that for circle infractions this should not be a thing, and perhaps the same for sector violations

I'm not gonna make this ticket super detailed since I know you know the codebase and what to do

owens3364:
Closes #33 

Note that the confusion about the math was not actually a math bug. The frontend was sending invalid requests when trying to change the throw type, so regardless of what throw type was selected, the fake throw data was always for a shot put field.

This was fixed by
- making the frontend lowercase the throw type in its POST requests to /api/throw-type , so it sends valid requests
- Adding better error logging on the frontend
- Not letting the frontend update its throw type state unless that POST request succeeds

### DIFF
--- a/backend/src/throws/infractions.rs
+++ b/backend/src/throws/infractions.rs
@@ -7,10 +7,3 @@ pub enum InfractionType {
     RightSector,
     Circle,
 }
-
-#[derive(Debug, Clone, Serialize)]
-pub struct Infraction {
-    #[serde(rename = "type")]
-    pub infraction_type: InfractionType,
-    pub confidence: f32,
-}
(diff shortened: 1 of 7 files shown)


luke-venk (approved):
Silly mistake by me regarding the invalid non-lowercase POST request. Great catch!

PR #64, opened 2026-04-26
Detect the start of the throw with CV

rx9933:
- Includes functions to detect the start of the throw. This is based on a jerk calculation in pixel space. 
- Acceleration is calculated with standard FD, then smoothed with a convolutional filter. To get jerk, a simple first order difference is implemented.
- Currently, for throw 0, there is a person walking across the screen, which CV tracks; so, the start detection is a bit late/incorrect for throw 0. 
- There are about 50 frames for each video (long 20+ second video) so I'm hoping that's enough for Triangulation, for now.
- You can adjust the jerk threshold separately for throw 0 vs throw 1, if needed

### DIFF
--- a/analysis/opencv_approach/detect_start.py
+++ b/analysis/opencv_approach/detect_start.py
@@ -1,0 +1,733 @@
+import cv2
+import numpy as np
+import matplotlib.pyplot as plt
+from pathlib import Path
+from scipy.interpolate import CubicSpline
+from tuned_cv1k import extract_trajectory_from_video
+
+plt.rcParams['font.size'] = 18
+def _natural_cubic_spline_second_derivatives(t, y):
+    n = len(t)
+    if n < 3:
+        return np.zeros(n, dtype=float)
+
+    h = np.diff(t)
+    alpha = np.zeros(n, dtype=float)
+    alpha[1:-1] = (3.0 / h[1:]) * (y[2:] - y[1:-1]) - (3.0 / h[:-1]) * (y[1:-1] - y[:-2])
+
+    l = np.ones(n, dtype=float)
+    mu = np.zeros(n, dtype=float)
+    z = np.zeros(n, dtype=float)
+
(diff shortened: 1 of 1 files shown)


analysis/opencv_approach/detect_start.py
```diff
@@ -0,0 +1,733 @@
+import cv2
```
luke-venk, line 1: Hey did you mean to add all these rust changes in `backend` to this PR, or just the one `detect_start.py` file?

If you meant to just want us to review the one Python file, please remove everything else from this PR

luke-venk (changes requested):

analysis/opencv_approach/detect_start.py
```diff
@@ -0,0 +1,733 @@
...
+    # colorbars (one per throw)
+    cbar1 = plt.colorbar(sc1, fraction=0.02, pad=0.02)
+    cbar1.set_label("Throw 1 time (start → end)")
+
+    cbar2 = plt.colorbar(sc2, fraction=0.02, pad=0.08)
+    cbar2.set_label("Throw 2 time (start → end)")
+
+    plt.legend()
+    plt.savefig('a1a2.png')
+    plt.show()
```
owens3364: Nit: add newline to end of file

owens3364 (changes requested):
How can I test/verify this? What should I run to see the results of your code?

rx9933:
You can run this file with two mp4 videos, for example:

- download throw0_med and throw1_med from [Box](https://utexas.box.com/s/nwhmif56sx9ioo3d2ixu7sgvoy4nrnxu)
- place in same directory as detect_start.py and the cv pipeline of choice
- run detect_start.py (ensure that the mp4 file names are edited if needed in detect_start).

Note that for one of the newer videos (throw_), a person walks across the frame, which ruins/hurts CV performance.

Issue #32, opened 2026-04-12
Return Decision From Math Triangulation to Frontend


luke-venk:
## Description
Once [Yash's triangulation PR](https://github.com/luke-venk/mjolnir/pull/14) is merged in, and #28 is complete, we need to convert whatever the outputs of the math triangulation function are (pretty sure these are pixel coordinates or something) to determine whether the object is in bounds or a sector violation. Confirm with @yashj1579 to figure out what outputs are, but this shouldn't be very difficult.

Once we have that, convert these outputs into a `ThrowAnalysisResponse` and send it to frontend. See below information.

## Relevant Files
- `backend/src/main.dev.rs`: Look at "real" `main()` (see README.bazel.md "Fake vs. Real" section for more info)
- `backend/src/main.prod.rs`: Look at "real" `main()` (see README.bazel.md "Fake vs. Real" section for more info)
- `backend/src/math_triangulation/math_triangulation.rs`: Figure out what outputs of this is
- `backend/src/server/server.rs`: The `get_throw_results` function, if the source is camera and not simulated data, should call whatever function you write (instead of `simulate_throw_event`) to return a `ThrowAnalysisResponse` type with the information you gathered from the math triangulation

Issue #28, opened 2026-04-11
Aggregate CV Pipeline Outputs into Math Triangulation


luke-venk:
## Description
Once [Yash's triangulation PR](https://github.com/luke-venk/mjolnir/pull/14) is merged in, we need to implement the plumbing between the 2 CV pipelines to connect to the math triangulation function. Whatever parameters the math triangulation needs, will need to be extracted from the output frames from the last stage of the CV pipelines.

## Work Needed
While our system is meant to use PTP to synchronize when the cameras take frames, acquisition for each camera isn't guaranteed (and never will) to start at the exact same time, despite it being multithreaded. That means that frame1 for the left camera won't correspond to frame1 for the right camera.

For example, look at [Luke's slides in the Beta presentation](https://docs.google.com/presentation/d/1N7rILkqB8jNnXUudb2pAsoiOmu9s-HGCMLUBd4UgWkA/edit?slide=id.g3de7effcc53_0_217#slide=id.g3de7effcc53_0_217). These 2 frames correspond to the same moment but are 15 frames apart.

Here is the metadata for frame 82 of the right camera.
```
{
  "camera_id": "Lucid Vision Labs-ATP124S-M-224300917",
  "frame_index": 82,
  "width": 4096,
  "height": 3000,
  "payload_bytes": 12288000,
  "system_timestamp_ns": 1775771934342128000,
  "buffer_timestamp_ns": 3707524399352,
  "frame_id": 83
}
```

Here is the metadata for frame 67 of the left camera.
```
{
  "camera_id": "Lucid Vision Labs-ATP124S-M-242700635",
  "frame_index": 67,
  "width": 4096,
  "height": 3000,
  "payload_bytes": 12288000,
  "system_timestamp_ns": 1775771934343718000,
  "buffer_timestamp_ns": 3707526308456,
  "frame_id": 68
}
```

It is up to you to figure out how to do this, but my first instinct is to use the crossbeam channel senders for the last CV stage (right now this is called `tx_stage5` in `backend/src/pipeline/pipeline.rs`, and have them send to the same crossbeam channel receiver to whatever you decide will aggregate the results. There should only be one output receiver so that will need to be passed into the Pipeline struct, and it will get the output frames from both camera pipelines.

[...]

PR #69, opened 2026-04-27
Return Throw Decision From Triangulation to Frontend (#32)

RushilRandhar:
## Summary

Adds a new `throws::decision` module that converts the optimized 3D trajectory from `math_triangulation::optimize_trajectory` into a `ThrowAnalysisResponse` for the frontend.

This is the **pure decision logic** half of #32. The wiring (server.rs, main.{dev,prod}.rs) depends on the channel type emitted by #28's aggregator, which is still in flight — see https://github.com/luke-venk/mjolnir/issues/28#issuecomment-4324017708 where I asked @MaxWiesenfeld and @yashj1579 to confirm the contract. Will add the wiring in a follow-up commit on this branch once #28 lands.

Opening as **draft** for that reason.

## What's in this PR

- **`landing_from_trajectory`** — picks the last 3D point as the (x, y) landing.
- **`compute_distance`** — distance from inside edge of throwing circle. Skips the radius subtraction for javelin since its field-dimension entry is the runway length, not a circle.
- **`classify_sector`** — returns `LeftSector` / `RightSector` / `None` based on the angle from +x axis vs. the half sector angle.
- **`build_throw_response`** — assembles full response. Returns `None` when LM failed or trajectory is empty (caller decides fallback).

Coordinate frame is documented as a TODO awaiting confirmation from @yashj1579: origin at circle center, +x down sector centerline.

## Open questions for reviewers

1. Once #28 settles the channel type, should `build_throw_response` accept the channel struct directly, or should the caller destructure it? Currently signature takes individual params for testability.
2. Image URLs are passed in by the caller. Where do these come from for the real pipeline — saved frame paths from the recording writer, or placeholder URLs for MVP? Will follow up with @luke-venk.

## Test plan

- [x] 18 unit tests covering boundary cases, javelin-specific distance, all infraction permutations
- [x] `bazel test //backend:tests` — all passing
- [x] `bazel build //backend/...` — all 18 targets build
- [ ] End-to-end test pending #28 + wiring follow-up

```
running 18 tests
test throws::decision::tests::landing_returns_last_trajectory_point ... ok
test throws::decision::tests::landing_empty_trajectory_returns_none ... ok
test throws::decision::tests::distance_subtracts_circle_radius_for_throwing_events ... ok
test throws::decision::tests::distance_uses_euclidean_norm ... ok
test throws::decision::tests::distance_negative_y_does_not_break_computation ... ok
test throws::decision::tests::distance_for_javelin_does_not_subtract_runway_length ... ok
test throws::decision::tests::sector_in_bounds_returns_none ... ok
test throws::decision::tests::sector_straight_down_middle_is_in_bounds ... ok
test throws::decision::tests::sector_left_violation_for_positive_y ... ok
test throws::decision::tests::sector_right_violation_for_negative_y ... ok
test throws::decision::tests::sector_javelin_uses_narrower_angle ... ok
test throws::decision::tests::build_response_in_bounds_no_circle_infraction ... ok
test throws::decision::tests::build_response_sector_violation_hides_landing_point ... ok
test throws::decision::tests::build_response_circle_infraction_passes_through ... ok
test throws::decision::tests::build_response_sector_and_circle_both_reported ... ok
test throws::decision::tests::build_response_returns_none_when_lm_failed ... ok
test throws::decision::tests::build_response_returns_none_for_empty_trajectory ... ok
test throws::decision::tests::timestamp_ns_round_trips_to_rfc3339 ... ok

test result: ok. 18 passed; 0 failed
```

Refs #32, blocked on #28 for full close-out.


backend/src/throws/decision.rs
```diff
@@ -0,0 +1,326 @@
+// Convert the output of `math_triangulation::optimize_trajectory` into a
+// `ThrowAnalysisResponse` that can be returned to the frontend.
+//
+// The triangulation step returns a 3D trajectory in world coordinates. This
+// module is responsible for picking the landing point, computing the distance
+// from the inside edge of the throwing circle, deciding whether the throw is a
+// sector violation, and assembling the final response.
+//
+// Coordinate frame:
```
luke-venk: @yashj1579 does this coordinate frame look good to you?

backend/src/throws/decision.rs
```diff
@@ -0,0 +1,326 @@
...
+use uuid::Uuid;
+
+// Pick the landing point from the optimized trajectory.
+//
+// The trajectory is a sequence of 3D positions sampled at `dt` intervals.
+// The landing point is the last sample (the trajectory is truncated at impact
+// during optimization). Returns `None` if the trajectory is empty.
+pub fn landing_from_trajectory(trajectory: &[Vector3<f64>]) -> Option<(f32, f32)> {
+    let last = trajectory.last()?;
+    Some((last[0] as f32, last[1] as f32))
```
luke-venk, line 39: In what case would the `trajectory` argument ever not have values, which could cause the function to return `None`?

backend/src/throws/decision.rs
```diff
@@ -0,0 +1,326 @@
...
+}
+
+// Distance from the inside edge of the throwing circle to the landing point,
+// in meters. The throwing circle is centered at the origin in our coordinate
+// frame, so the distance is `||(x, y)|| - circle_radius`.
```
luke-venk: @yashj1579 this looks good to you right?

backend/src/throws/decision.rs
```diff
@@ -0,0 +1,326 @@
...
+// is from the inside edge of the circle. For javelin, the "circle" field
+// dimension represents the runway length, not a circle radius, so the
+// subtraction is omitted.
+pub fn compute_distance(landing_xy: (f32, f32), throw_type: ThrowType) -> f32 {
+    let (x, y) = landing_xy;
+    let raw = (x * x + y * y).sqrt();
+    match throw_type {
+        ThrowType::Javelin => raw,
+        _ => {
+            let (circle_diameter, _, _) = get_field_dimensions(throw_type);
```
luke-venk, line 57: Is this math right? I lowkey thought trig was needed to figure out the distance from the edge as opposed to the origin. Maybe not though. @owens3364

backend/src/throws/decision.rs
```diff
@@ -0,0 +1,326 @@
...
+use chrono::{DateTime, Utc};
+use nalgebra::Vector3;
+use uuid::Uuid;
+
+// Pick the landing point from the optimized trajectory.
+//
+// The trajectory is a sequence of 3D positions sampled at `dt` intervals.
+// The landing point is the last sample (the trajectory is truncated at impact
+// during optimization). Returns `None` if the trajectory is empty.
+pub fn landing_from_trajectory(trajectory: &[Vector3<f64>]) -> Option<(f32, f32)> {
```
luke-venk: nit: make function private?

backend/src/throws/decision.rs
```diff
@@ -0,0 +1,326 @@
...
+// Distance from the inside edge of the throwing circle to the landing point,
+// in meters. The throwing circle is centered at the origin in our coordinate
+// frame, so the distance is `||(x, y)|| - circle_radius`.
+//
+// World athletics rules measure from the inside edge of the stop board, but
+// for shot put / discus / hammer the circle is symmetric and the measurement
+// is from the inside edge of the circle. For javelin, the "circle" field
+// dimension represents the runway length, not a circle radius, so the
+// subtraction is omitted.
+pub fn compute_distance(landing_xy: (f32, f32), throw_type: ThrowType) -> f32 {
```
luke-venk: nit: make function private?

backend/src/throws/decision.rs
```diff
@@ -0,0 +1,326 @@
...
+    }
+}
+
+// Decide whether the landing point is outside the legal sector and, if so,
+// which side it landed on. Returns `None` if the throw is in-bounds.
+//
+// The sector is symmetric about the +x axis with half-angle `sector_angle / 2`.
+// A throw with positive y past the half-angle is a `LeftSector` violation;
+// negative y past the half-angle is `RightSector`.
+pub fn classify_sector(landing_xy: (f32, f32), throw_type: ThrowType) -> Option<InfractionType> {
```
luke-venk: nit: make function private?

backend/src/throws/decision.rs
```diff
@@ -0,0 +1,326 @@
...
+}
+
+// Build the final response from the triangulation output and surrounding
+// context. `circle_infraction` comes from the touch sensor pipeline and is
+// provided by the caller; this module does not detect circle infractions.
+//
+// Returns `None` if the triangulation failed or produced no trajectory.
+// The caller decides what to do in that case (e.g., return 503, fall back
+// to the previous response, etc).
+pub fn build_throw_response(
```
luke-venk, line 118: Nice, so this is the only public function that would actually be called elsewhere right? I think that your next PR will actually be calling this function whenever triangulation is complete

backend/src/throws/decision.rs
```diff
@@ -0,0 +1,326 @@
...
+    trajectory: &[Vector3<f64>],
+    lm_success: bool,
+    throw_type: ThrowType,
+    impact_timestamp_ns: u64,
+    circle_infraction: Option<InfractionType>,
+    image_urls: Vec<String>,
+) -> Option<ThrowAnalysisResponse> {
+    if !lm_success {
+        return None;
```
luke-venk: Few things:
1. What is `lm_success`, is this implying that CV could either fail or not? Also, how would we know if `lm_success` was true or not?
2. If a failure did happen, do we just want to return `None`, or should we return a `ThrowResponse` still with some indication of an error

@owens3364

backend/src/throws/decision.rs
```diff
@@ -0,0 +1,326 @@
...
+        throw_type,
+        distance_m: compute_distance(landing_xy, throw_type),
+        infractions,
+        images: image_urls,
+        landing_point_x_y,
+    })
+}
+
+#[cfg(test)]
+mod tests {
```
luke-venk, line 160: nice unit tests!

luke-venk (changes requested):

backend/src/throws/decision.rs
```diff
@@ -0,0 +1,326 @@
...
+// Returns `None` if the triangulation failed or produced no trajectory.
+// The caller decides what to do in that case (e.g., return 503, fall back
+// to the previous response, etc).
+pub fn build_throw_response(
+    trajectory: &[Vector3<f64>],
+    lm_success: bool,
+    throw_type: ThrowType,
+    impact_timestamp_ns: u64,
+    circle_infraction: Option<InfractionType>,
+    image_urls: Vec<String>,
```
luke-venk, line 122: Good call on the URLs. Can I ask you to handle this? I will make a ticket for this #75

luke-venk:
There are also merge conflicts

backend/src/throws/decision.rs
RushilRandhar: Done in c398132 — also made `compute_distance` and `classify_sector` private for the same reason. `build_throw_response` is now the only `pub fn` in the module.

backend/src/throws/decision.rs
RushilRandhar: Done in c398132.

backend/src/throws/decision.rs
RushilRandhar: Done in c398132.

backend/src/throws/decision.rs
RushilRandhar: In practice the trajectory is empty only if the upstream LM-call layer (issue #80) emits a `TriangulationOutput` with `triangulation_succeeded == true` but no samples — which shouldn't happen with the current `optimize_trajectory` semantics, but the type is `Vec<Vector3<f64>>` so the API stays defensive. `build_throw_response` already short-circuits on `triangulation_succeeded == false`; this is the secondary defense if the upstream contract drifts. Documented inline in c398132.

backend/src/throws/decision.rs
RushilRandhar: @yashj1579 ack confirmed in commit 9bb8dc5 ('docs: confirm coordinate frame origin at circle center') — dropped the TODO. Resolving this thread; flag if anything's drifted since.

backend/src/throws/decision.rs
RushilRandhar: @yashj1579 — need explicit ack on the sector-violation logic too. Specifically: `y > 0` past the half-angle = `LeftSector`, `y < 0` past the half-angle = `RightSector`. Does that match the convention you used in calibration, or should it be the other way around?

backend/src/throws/decision.rs
RushilRandhar: @owens3364 to double-check, but I think the math is right given the coordinate frame Yash confirmed (origin at circle center, +x down sector centerline):

- Landing point is (x, y) in meters from the origin.
- `||(x, y)|| = sqrt(x² + y²)` is the Euclidean distance from origin to landing point.
- Origin = circle center, so `||(x, y)|| - circle_radius` = distance from inside edge of circle to landing point. (Subtract the radius, not the diameter.)

No trig is needed because we're already in 2D Cartesian after the trajectory output. Trig would only come in if (a) the origin was somewhere else (e.g. the inside edge of the stop board on shot put, which isn't centered on the circle), or (b) we needed to project onto a different axis. Neither applies here.

For javelin we skip the subtraction because the "circle" entry in `get_field_dimensions` is the runway length (16 m), not a circle radius — would produce a wildly wrong distance otherwise (test `distance_for_javelin_does_not_subtract_runway_length` covers that).

Owen, sanity check?

backend/src/throws/decision.rs
RushilRandhar: Yes — `build_throw_response` is now the only `pub fn` in the module after c398132 (the three helpers are all private now). The next caller will be #80 (the LM-call thread), which produces the `TriangulationOutput` and passes it into `build_throw_response`. Server-side wiring follows after that.

backend/src/throws/decision.rs
RushilRandhar: 1. **What is `lm_success`?** It's the Levenberg-Marquardt non-linear least-squares solver's convergence flag — the 4th tuple element returned from `math_triangulation::optimize_trajectory(...)` ([line 184 of `math_triangulation.rs`](https://github.com/luke-venk/mjolnir/blob/main/backend/src/math_triangulation/math_triangulation.rs#L184)). The LM optimizer iteratively refines the 3D trajectory to minimize reprojection error against the observed pixel tracks; if it doesn't converge within its iteration budget, `success == false` and the resulting trajectory is whatever the last iterate happened to be — not trustworthy.

   Refactored in c398132 to take `&TriangulationOutput { trajectory, triangulation_succeeded, impact_timestamp_ns }` so the field is self-documenting. CV itself doesn't fail/not-fail in this sense — CV runs always; it's the post-CV LM refinement that may not converge.

2. **Failure mode — `None` vs partial response?** I went with `None` for two reasons. (a) Distance/sector classification on a non-converged trajectory would produce a confidently-wrong answer (e.g. a trajectory that diverged off-field would be reported as a record-breaking throw), worse than no answer. (b) Returning `None` keeps the policy at the call site: the server can decide whether to return 503, fall back to the previous response, show a "no valid throw" indicator, etc. — different stakeholders may want different UX. If we ever need a typed error variant (`Err("LM did not converge")` instead of `None`), that's a small change to the return type.

@owens3364 do you have a strong preference?

backend/src/throws/decision.rs
RushilRandhar: 🙏

backend/src/throws/decision.rs
RushilRandhar: On it — will pick up #75 once this PR lands. Image URLs stay as a caller-passed param in this PR; #75 will make them real (saved frame paths from the recording writer).

RushilRandhar:
Pushed c398132 + d0f72c7 (merge of main).

## Summary of what changed

- **Merge conflict resolved.** `backend/BUILD.bazel`: `backend_lib_all_fake` deps list took both `nalgebra` (this PR's addition for decision logic) and `opencv` (main's addition from #47 downsampling).

- **All inline review comments addressed.** Replies posted on each thread. Highlights:
  - `landing_from_trajectory`, `compute_distance`, `classify_sector` are now private (3 nits).
  - `build_throw_response` now takes a `&TriangulationOutput` struct instead of separate `trajectory + lm_success` params. Cleaner, self-documenting.
  - `lm_success` field renamed to `triangulation_succeeded` and documented as the LM convergence flag from `optimize_trajectory`.

- **Pinned the contract for the missing middle step.** This PR is decision logic; the LM-call layer that consumes Max's `OptimizeTrajectoryInput` and produces `TriangulationOutput` is currently nobody's ticket. Filed #80 for it (unassigned for now). #80 also depends on #74 stabilizing and a calibration-loading mechanism. Left a comment on PR #74 flagging that `OptimizeTrajectoryInput` will need an `impact_timestamp_ns` field for #80 to populate the response timestamp.

- **`TriangulationOutput` lives in a new `crate::triangulation` module** rather than next to `optimize_trajectory` because Yash's `math_triangulation` module is currently orphaned (not declared in `lib.rs` and has pre-existing test compile issues that surfaced when I tried to wire it in). Hooking it up is out of scope for this PR; the new `triangulation` module is the long-term home for #80's work too.

## Tests

`bazel test //backend:tests` — **33 passed, 0 failed.** All 18 of the original decision tests still pass after the refactor, plus existing tests on main continue to pass.

## Outstanding asks

- @yashj1579: confirmed coordinate frame in 9bb8dc5; need an explicit ack on the sector-violation logic (left/right convention).
- @owens3364: sanity check on the distance-formula reasoning and whether `None`-vs-typed-error is the right failure mode.

Status stays draft until #80 (LM-call) and Max's #74 land, since end-to-end execution is gated on those. The decision logic itself is reviewable independently.

backend/src/throws/decision.rs
RushilRandhar: https://coachingyoungathletes.com/2022/07/21/how-to-measure-a-shot-put-throw/

backend/src/triangulation/triangulation_output.rs
```diff
@@ -0,0 +1,27 @@
+use nalgebra::Vector3;
+
+// Output contract that downstream consumers (decision logic, server) take from
```
owens3364: Can you change all of these from regular comments (two slashes) to rustdoc comments (three slashes)? That way it'll show up in the editor/ide/vs code

backend/src/triangulation/triangulation_output.rs
```diff
@@ -0,0 +1,27 @@
...
+// `math_triangulation::optimize_trajectory` (see issue #80) by packaging its
+// return tuple together with the impact timestamp from the matched-pair
+// window.
+//
+// Covariance and drag are intentionally not included here — neither is
+// consumed by the decision logic today. They can be added later if a consumer
+// needs them.
+#[derive(Debug, Clone)]
+pub struct TriangulationOutput {
+    // 3D positions sampled at `dt` intervals from `optimize_trajectory`.
```
owens3364: Same thing here

backend/src/math_triangulation/triangulation_output.rs
```diff
@@ -0,0 +1,27 @@
...
+    pub trajectory: Vec<Vector3<f64>>,
+    // Whether the Levenberg-Marquardt non-linear solver in
+    // `optimize_trajectory` converged. False means the trajectory is not
+    // usable; consumers should surface a "no valid throw" response upstream
+    // rather than running distance/sector classification on garbage output.
+    pub triangulation_succeeded: bool,
+    // Buffer timestamp of the impact frame, propagated end-to-end so the
+    // decision logic can populate `frame_timestamp_from_camera_microseconds`
+    // on the response.
+    pub impact_timestamp_ns: u64,
```
owens3364, line 33: Clarification - is this the buffer timestamp of the frame, or the interpolated estimate of when it actually impacted the ground?

It is rare that we would capture a frame of the exact impact time. Once we have a polynomial fit, we can solve for y/z = 0 to find impact time.

Can you expand on exactly what this is and why it's necessary/useful?

This might be the right property I just want to understand why

backend/src/triangulation/mod.rs
```diff
@@ -0,0 +1,16 @@
+// Output-side contract for the triangulation stage.
```
owens3364: Use rustdoc comments

backend/src/throws/decision.rs
```diff
@@ -0,0 +1,332 @@
...
+fn landing_from_trajectory(trajectory: &[Vector3<f64>]) -> Option<(f32, f32)> {
+    let last = trajectory.last()?;
+    Some((last[0] as f32, last[1] as f32))
+}
+
+// Distance from the inside edge of the throwing circle to the landing point,
+// in meters. The throwing circle is centered at the origin in our coordinate
+// frame, so the distance is `||(x, y)|| - circle_radius`.
+//
+// World athletics rules measure from the inside edge of the stop board, but
```
owens3364: Make this a rustdoc please

backend/src/throws/decision.rs
```diff
@@ -0,0 +1,332 @@
...
+    } else {
+        InfractionType::RightSector
+    })
+}
+
+// Convert a buffer timestamp in nanoseconds to the ISO-8601 string format
+// the frontend expects. The schema field is named
+// `frame_timestamp_from_camera_microseconds` for historical reasons but the
+// value is an RFC3339 timestamp string (see `simulate_throw_event`).
+fn timestamp_ns_to_string(impact_timestamp_ns: u64) -> String {
```
owens3364: Unless the `impact_timestamp_ns` is already referenced to the unix epoch, this will generate vastly incorrect datetimes.

Can you clarify either
1) That I'm missing something and wrong
2) That somewhere upstream a caller is turning the camera ptp timestamps into unix-epoch-referenced timestamps

Or if not, this should be handled differently or explicitly require a unix epoch referenced input timestamp.

backend/src/throws/decision.rs
owens3364: Instead of using `Option` to account for errors we should use Rust's builtin `Result` type. `Result<ThrowAnalysisResponse, String>` would work fine here IMO with the example error message you already have written!

owens3364 (changes requested):

backend/src/triangulation/triangulation_output.rs
RushilRandhar: Done in 83817b2 — converted to `///` rustdoc throughout `triangulation_output.rs`.

backend/src/triangulation/triangulation_output.rs
RushilRandhar: Done in 83817b2.

backend/src/triangulation/mod.rs
RushilRandhar: Done in 83817b2 — module-level doc now uses `//!`.

backend/src/throws/decision.rs
RushilRandhar: Done in 83817b2 — converted all type/function docs in `decision.rs` to `///` and the module header to `//!`.

backend/src/math_triangulation/triangulation_output.rs
RushilRandhar: Good catch — you're right that this should be the interpolated impact instant, not a captured frame's buffer timestamp. Updated the doc in 83817b2 to make that explicit:

> Interpolated impact instant in **Unix-epoch nanoseconds**.
>
> Cameras only sample at the configured frame rate, so we rarely capture a frame at the exact moment of impact. The LM-call layer (#80) is responsible for solving the trajectory polynomial for the y/z = 0 crossing and producing the sub-frame-precision timestamp. The decision logic uses this to populate \`frame_timestamp_from_camera_microseconds\` on the response.

This is part of the upstream contract — `TriangulationOutput` is just the data shape; the actual interpolation happens in #80. PR #69 is just the consumer side here.

Why it's necessary: the frontend currently displays \`frame_timestamp_from_camera_microseconds\` (misnomer — it's an RFC3339 string in the schema). With sub-frame precision we get a meaningful "when did this throw land" stamp; with frame-buffer precision the value would jitter by ±33ms at 30Hz and miss the actual landing instant. Sound reasonable?

backend/src/throws/decision.rs
RushilRandhar: You're right to flag this — the cameras' raw \`buffer_timestamp_ns\` is PTP (TAI epoch / device-specific), not Unix epoch. \`from_timestamp\` would silently produce nonsense.

Updated in 83817b2 to make the precondition explicit. The function now takes \`impact_unix_timestamp_ns\`, and its doc says:

> **Precondition:** \`impact_unix_timestamp_ns\` must be referenced to the Unix epoch. The cameras' raw \`buffer_timestamp_ns\` values are PTP timestamps in their own clock domain; converting from camera/PTP time to Unix time is upstream's responsibility (tracked in #76). Passing in raw PTP nanoseconds will produce nonsense datetimes.

The PTP→Unix conversion isn't this PR's scope — issue #76 ("Find Difference Between Local PC Time and Camera PTP Time") is tracking that work and Eloghosa's assigned. \`TriangulationOutput::impact_timestamp_ns\` is also doc'd as Unix-epoch ns now, so the contract is consistent end-to-end.

backend/src/throws/decision.rs
RushilRandhar: Done in 83817b2. \`build_throw_response\` now returns \`Result<ThrowAnalysisResponse, String>\`. Two distinct error messages:

- \`Err(\"Levenberg-Marquardt optimization did not converge\")\` when \`triangulation_succeeded == false\`
- \`Err(\"Triangulation produced an empty trajectory\")\` when the trajectory is empty after a successful solve

Renamed the two relevant tests to \`build_response_returns_err_*\` and updated their assertions accordingly.

RushilRandhar: Done in 83817b2 — switched to `Result<ThrowAnalysisResponse, String>`. Two distinct error messages:

- `Err("Levenberg-Marquardt optimization did not converge")` when `triangulation_succeeded == false`
- `Err("Triangulation produced an empty trajectory")` when the trajectory is empty after a successful solve

(I also replied on the parallel thread Luke started about `lm_success`; mirroring here so this thread is explicitly resolved too.)

backend/src/triangulation/mod.rs
```diff
@@ -0,0 +1,16 @@
+//! Output-side contract for the triangulation stage.
```
luke-venk: could you pls move to the already existing `math_triangulation`?

luke-venk (dismissed):

owens3364 (approved):

PR #71, opened 2026-04-28
better ptp cli settings

owens3364:


backend/src/bin/record_from_both_cameras.rs
```diff
@@ -134,7 +134,11 @@ pub fn main() {
             Some(addr),
             Some(configuration_barrier_1),
             Some(acquisition_barrier_1),
-            Some(ptp_config_1),
+            if args.common_args.enable_ptp {
```
luke-venk: nit: I think it'd be nicer to define `ptp_config1` outside the function call

```
let ptp_config = if args.common_args.enable_ptp {
    Some(ptp_config_1)
} else {
    None
};
```

and pass it in

backend/src/bin/record_from_both_cameras.rs
```diff
@@ -151,7 +155,11 @@ pub fn main() {
             None,
             Some(configuration_barrier_2),
             Some(acquisition_barrier_2),
-            Some(ptp_config_2),
+            if args.common_args.enable_ptp {
```
luke-venk: same as https://github.com/luke-venk/mjolnir/pull/71/changes#r3151189918

luke-venk (changes requested):

luke-venk (approved):

PR #74, opened 2026-04-28
Aggregate CV Pipeline Outputs

### LINKED ISSUE #28: Aggregate CV Pipeline Outputs into Math Triangulation

MaxWiesenfeld:
## Justification

This pull request satisfies #28 whose goal was to allow us to bring information from the cameras through CV to optimize_trajectory. 

## Implementation

1. We have enhanced the Context data structure to hold optional cx, cy and detected bool flag--outputs of Arushi's current Python `contour` implementation
3. After running through pipeline--at contour.rs--Context is enriched and optional fields are populated
4. Strip Frame into ContourOutput (slimmed metadata structure which stores no raw image data)
5. Frames are cleared of CV steps
6. Aggregator consumes ContourOutputs
7. Aggregator uses FPS and closest frames to stitch together matched pairs
8. Trajectory input collector consumes matched pairs
9. Collector finds dt from median timestamp delta between frames
10. Collector has option to optimize/filter by usable Context by filtering out pairs with 1 or more missing `detected=True`
11. Collector creates pixel over time vector
12. (To be implemented): Collector(?) owns P list of config variable and functions / remaining inputs 
13. (To be implemented): Collector feeds to optimize_trajectory

### Note: This also required adding --left-camera and --right-camera cli args which must be populated with a camera's name


backend/src/camera/cli_real.rs
```diff
@@ -0,0 +1,15 @@
+use clap::Parser;
+
+/// Command line arguments for running the real-camera CV backend pipelines.
+#[derive(Parser, Debug, Clone)]
+#[command(name = "cv_backend_camera_args")]
+#[command(about = "Runs the real-camera CV backend pipelines with explicit left/right camera names.")]
+pub struct CvBackendCameraArgs {
+    /// Camera identifier for the left-field camera pipeline.
+    #[arg(long = "left-camera", required = true)]
```
luke-venk: How would this actually be used, can you please update `README.cameras.md`?

backend/src/camera_ingest/camera_ingest.rs
```diff
@@ -1,9 +1,17 @@
-use crate::pipeline::Frame;
+use crate::pipeline::{CameraId, Context, Frame};
 
 // Ingests frames from the cameras using the GigEVision API, and enqueues
 // the frames into the camera ingest sender for the camera's pipeline to
 // begin processing.
-pub fn ingest_frames(_tx: crossbeam::channel::Sender<Frame>) {
-    // TODO(#3): Implement Camera Ingest.
-    todo!();
+pub fn ingest_frames(
```
luke-venk, line 6: I could be missing something, but did you have to update this function to get your work done? @grao23 is handling ingest and I'd imagine this has no impact on your immediate PR

if it has no impact can you please revert changes to ingest as to not conflict with Gautam?

backend/src/pipeline/frame.rs
```diff
@@ -92,19 +132,26 @@ pub enum CameraId {
...
+#[derive(Debug, Clone, Copy, PartialEq)]
 pub struct Context {
     camera_id: CameraId,
 
     /// The timestamp given to the buffer by the camera (i.e. buffer.timestamp()).
     camera_buffer_timestamp: u64,
+
+    detected: Option<bool>,
+    cx: Option<f64>,
+    cy: Option<f64>,
```
luke-venk: Why not just make this an `Option<(f64, f64)>` so they're more coupled? Then you could make your getter/setter cleaner

backend/src/computer_vision/forward_downsampled_copy.rs
```diff
@@ -4,9 +4,8 @@
 // frames from color to grayscale, but our cameras are already monochrome.
```
luke-venk, line 4: It's cool that formatter was run but could you please undo the deletion of my comments lol

backend/src/pipeline/pipeline.rs
```diff
@@ -16,26 +17,30 @@ impl Pipeline {
...
+            ingest_frames(camera_id, camera_name, tx_ingest);
         });
 
         // Stage 1: Undistortion.
         let handle_stage1 = PipelineStage::new(rx_stage1, tx_stage1, undistortion).spawn();
-
         // Stage 2: Forward Downsampled Copy.
         let handle_stage2 =
             PipelineStage::new(rx_stage2, tx_stage2, forward_downsampled_copy).spawn();
+        let (tx_stage4, rx_output) = bounded::<Frame>(capacity_per_channel);
```
luke-venk: lol why did this get randomly moved from line 25, please move back

backend/src/pipeline/pipeline.rs
luke-venk: nit: rename to `camera_id` (i.e. remove underscore now that it's no longer unused)

backend/src/pipeline/pipeline_stage.rs
```diff
@@ -1,31 +1,34 @@
-use crate::pipeline::Frame;
 use crossbeam::channel::{Receiver, Sender};
 use std::thread;
 
-pub struct PipelineStage<F>
+pub struct PipelineStage<In, Out, F>
```
luke-venk: What is the point in making the pipeline stage able to handle more than just frames? Are we not just sending frames through?

backend/src/schemas/contour_output.rs
```diff
@@ -0,0 +1,174 @@
+use crate::pipeline::{CameraId, Frame};
```
luke-venk: I deleted this directory in a previous PR. Given that it literally only holds one file, can we move to `pipeline/` directory?

backend/src/pipeline/contour_output.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub fn center_px(&self) -> Option<&PixelCenter> {
+        self.center_px.as_ref()
+    }
+
+    pub fn detected(&self) -> bool {
+        self.detected
+    }
+}
+
+impl From<Frame> for ContourOutput {
```
luke-venk: Also per our [convo last night](https://github.com/luke-venk/mjolnir/issues/28#issuecomment-4332252304), I thought we agreed on keeping the original `Frame` struct so we could refer to the information needed. Am I missing something? I thought we needed the original high resolution information to do triangulation, but I could be mistaken

@owens3364

backend/src/aggregation_coordinator.rs
```diff
@@ -0,0 +1,168 @@
+use crate::schemas::MatchedContourPair;
```
luke-venk: Can you organize this file into `backend/src/aggregator` or `backend/src/pipeline/aggregator` (whichever one you think makes most sense)?

Could you also please add a docstring summarizing this file?

backend/src/matched_contour_pair_aggregator.rs
```diff
@@ -0,0 +1,50 @@
+use crate::pipeline::CameraId;
```
luke-venk: Same as https://github.com/luke-venk/mjolnir/pull/74/changes#r3155049146

backend/src/trajectory_input_collector.rs
```diff
@@ -0,0 +1,185 @@
+use crate::schemas::MatchedContourPair;
```
luke-venk: Same as https://github.com/luke-venk/mjolnir/pull/74/changes#r3155049146

backend/src/pipeline/pipeline.rs
```diff
@@ -16,26 +17,30 @@ impl Pipeline {
     // create the PipelineStages, spawn each stage, and return a Pipeline.
     // Calling this function automatically starts each of the pipeline
     // stages.
-    pub fn new(_camera_id: CameraId, capacity_per_channel: usize) -> Self {
+    pub fn new(
+        camera_id: CameraId,
+        camera_name: String,
```
luke-venk, line 18: I could be tripping but why is `camera_name` needed? From what I see it doesn't impact logic at all and just makes us need to supply command line arguments

backend/src/camera/cli_cv.rs
```diff
@@ -0,0 +1,15 @@
+use clap::Parser;
+
+/// Command line arguments for running the real-camera CV backend pipelines.
+#[derive(Parser, Debug, Clone)]
+#[command(name = "cv_backend_camera_args")]
+#[command(about = "Runs the real-camera CV backend pipelines with explicit left/right camera names.")]
+pub struct CvBackendCameraArgs {
```
luke-venk: See https://github.com/luke-venk/mjolnir/pull/74/changes#r3155068677

If this isn't needed can we delete this?

backend/src/main.dev.rs
```diff
@@ -40,10 +52,40 @@ async fn main() {
 #[cfg(feature = "real_cameras")]
 #[tokio::main]
 async fn main() {
+    let args: CvBackendCameraArgs = CvBackendCameraArgs::parse();
```
luke-venk: See https://github.com/luke-venk/mjolnir/pull/74/changes#r3155068677

If this isn't needed can we please delete this

backend/src/main.prod.rs
```diff
@@ -44,10 +56,40 @@ async fn main() {
 #[cfg(feature = "real_cameras")]
 #[tokio::main]
 async fn main() {
+    let args: CvBackendCameraArgs = CvBackendCameraArgs::parse();
```
luke-venk: Same as https://github.com/luke-venk/mjolnir/pull/74/changes#r3155074575

third_party/BUILD.bazel
```diff
@@ -1328,18 +1328,15 @@ cmake(
         "WITH_V4L": "OFF",
```
luke-venk, line 1501: Were these changes to this file needed for you to build? Pretty sure you undid changes I made so that build worked for me

luke-venk (changes requested):

backend/src/camera_ingest/camera_ingest.rs
MaxWiesenfeld: Pipeline's have to be named now by camera for us to track and forward to CV math and triangulation so this is necessary. You'll see I modified his TODO, but kept it just with new params

backend/src/pipeline/pipeline.rs
MaxWiesenfeld: It is needed. It assigns how we feed into the aggregator and input collector and eventually will determine how we pass the input pixels over time package to optimize_trajectory.

backend/src/pipeline/contour_output.rs
MaxWiesenfeld: My intention in this current PR was for us to shed the CV stages, but anticipated we would persist Frames somewhere else in the pipeline--they just weren't worth saving in the dedicated CV buffer (for aggregating inputs to optimize_trajectory) which is very cheap since it only holds (cx, cy), detected bool, timestamp, and camera pretty much. To me it is a branching point.

Let me know if yall disagree and want to get rid of ContourOutputs anyway and pass the whole frame through to aggregator and input collector.

backend/src/pipeline/pipeline.rs
luke-venk: Again, I could be missing something and don't understand yet, but like why would you aggregator need to know the IP address of the camera? I can see why it would need to know left or right to know which camera a frame came from, but why also the name specifically?

If I'm totally blind (I'm reading these super quickly since I'm reviewing everyone's PRs) can you just link where the camera name is used in the logic?

third_party/BUILD.bazel
MaxWiesenfeld: I wasn't able to a do a successful full build since some of Owen's changes yesterday. This was me trying to mess with things, but failing--will delete.

backend/src/camera_ingest/camera_ingest.rs
grao23: in my code the ingest_frames handle a part of live ingest, I have file which reads in folder and sorts and then ingests its called like replay_recorded_session.rs.

MaxWiesenfeld: Ok, maybe we should resync. Can you talk tomorrow morning?

grao23: yesi can, but I have handed this to Rushil to test further/tune.

backend/src/pipeline/pipeline.rs
```diff
@@ -1,13 +1,11 @@
...
 use crate::camera_ingest::ingest_frames;
 use crate::computer_vision::{contour, forward_downsampled_copy, mog2, undistortion};
-use crate::pipeline::{CameraId, Frame};
-use crate::schemas::ContourOutput;
+use crate::pipeline::{CameraId, ContourOutput, Frame};
 use crossbeam::channel::{Sender, bounded};
 use std::thread::{self, JoinHandle};
 
 pub struct Pipeline {
-    _camera_id: CameraId,
```
luke-venk, line 10: Does `Pipeline` no longer need a camera id?

luke-venk (changes requested):

RushilRandhar:
Hey @MaxWiesenfeld — flagging a gap that surfaced while planning #32's downstream consumer. Filed #80 to track filling it.

The decision logic in #32 needs to populate `frame_timestamp_from_camera_microseconds` on the response. That value should be the timestamp of the matched-impact frame (presumably the last `MatchedContourPair` consumed by the collector before the window closes). Right now `OptimizeTrajectoryInput` carries `pixels` and `dt_s` but not an impact timestamp, so when #80's owner wires the `optimize_trajectory(...)` call between this PR and #69's decision logic, the impact timestamp has nowhere to flow through.

Two options worth considering for when you address Luke's revisions:

1. Add a third field to `OptimizeTrajectoryInput`:
   ```rust
   pub struct OptimizeTrajectoryInput {
       pixels: Vec<Vec<Vector2<f64>>>,
       dt_s: f64,
       impact_timestamp_ns: u64,  // <- new; e.g. last matched pair's pair_timestamp_ns
   }
   ```
2. Leave the input struct alone and let #80's owner side-channel timestamps separately. Cleaner separation but adds a parallel channel for what's effectively trajectory-bound metadata.

(1) reads cleaner from #69's side. Does that work for you, or is there a reason to keep `OptimizeTrajectoryInput` minimal?

No action needed on this PR — just flagging so it's in your peripheral vision when you do the rework.

Refs #80

MaxWiesenfeld:
> Hey @MaxWiesenfeld — flagging a gap that surfaced while planning #32's downstream consumer. Filed #80 to track filling it.
> 
> The decision logic in #32 needs to populate `frame_timestamp_from_camera_microseconds` on the response. That value should be the timestamp of the matched-impact frame (presumably the last `MatchedContourPair` consumed by the collector before the window closes). Right now `OptimizeTrajectoryInput` carries `pixels` and `dt_s` but not an impact timestamp, so when #80's owner wires the `optimize_trajectory(...)` call between this PR and #69's decision logic, the impact timestamp has nowhere to flow through.
> 
> Two options worth considering for when you address Luke's revisions:
> 
> 1. Add a third field to `OptimizeTrajectoryInput`:
>    ```rust
>    pub struct OptimizeTrajectoryInput {
>        pixels: Vec<Vec<Vector2<f64>>>,
>        dt_s: f64,
>        impact_timestamp_ns: u64,  // <- new; e.g. last matched pair's pair_timestamp_ns
>    }
>    ```
> 2. Leave the input struct alone and let [Call optimize_trajectory between aggregator output and throw decision #80](https://github.com/luke-venk/mjolnir/issues/80)'s owner side-channel timestamps separately. Cleaner separation but adds a parallel channel for what's effectively trajectory-bound metadata.
> 
> (1) reads cleaner from #69's side. Does that work for you, or is there a reason to keep `OptimizeTrajectoryInput` minimal?
> 
> No action needed on this PR — just flagging so it's in your peripheral vision when you do the rework.
> 
> Refs #80

Noted and agreed--it shouldn't be in this PR. We haven't added Arushi's scripts for start and end detection yet and this PR is stubbed for those. In the end though, I am happy to extend OptimizeTrajectoryInput down the line, just shouldn't add unintuitive bloat to this PR.

backend/src/pipeline/pipeline.rs
MaxWiesenfeld: Take a look at let left_pixels and let right_pixels in `backend/src/aggregator/trajectory_input_collector.rs`. When building the pixel vector, we feed into it depending on which are left (first/[0]) or right (second/[1]).

backend/src/pipeline/pipeline.rs
MaxWiesenfeld: No, that was a verified unused remnant. After the constructor / passing camera ID to camera ingest, it can be shed from pipeline since it is then stored in Frame's Context

backend/src/aggregator/matched_contour_pair_aggregator.rs
```diff
@@ -0,0 +1,54 @@
...
+                    let right_ts = right.camera_buffer_timestamp();
+                    let delta = left_ts.abs_diff(right_ts);
+
+                    if delta <= expected_frame_interval_ns {
+                        let left_match = left_queue.pop_front().unwrap();
+                        let right_match = right_queue.pop_front().unwrap();
+                        let matched_pair = MatchedContourPair::new(left_match, right_match);
+                        let _ = matched_pair_tx.send(matched_pair);
+                    } else if left_ts < right_ts {
+                        let _ = left_queue.pop_front();
```
owens3364: Use `expect` here and two lines below it, no reason to ignore potentially odd results

backend/src/aggregator/matched_frame_pair_aggregator.rs
```diff
@@ -0,0 +1,54 @@
...
+                while let (Some(left), Some(right)) = (left_queue.front(), right_queue.front()) {
+                    let left_ts = left.camera_buffer_timestamp();
+                    let right_ts = right.camera_buffer_timestamp();
+                    let delta = left_ts.abs_diff(right_ts);
+
+                    if delta <= expected_frame_interval_ns {
+                        let left_match = left_queue.pop_front().unwrap();
+                        let right_match = right_queue.pop_front().unwrap();
+                        let matched_pair = MatchedContourPair::new(left_match, right_match);
+                        let _ = matched_pair_tx.send(matched_pair);
```
owens3364, line 46: Use `expect` here or exit the loop but don't ignore transmission send failures

backend/src/pipeline/contour_output.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    pub fn cy_px(&self) -> f64 {
+        self.cy_px
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct ContourOutput {
+    camera_id: CameraId,
```
owens3364: these data look familiar, v similar to the contents of `Context`

Can we cut dome on some of these properties?

I'm not saying it's the end of the world but it does feel odd to have the same properties in two places

backend/src/pipeline/contour_output.rs
owens3364: @luke-venk correct

backend/src/pipeline/contour_output.rs
```diff
@@ -0,0 +1,174 @@
...
+            frame.context().camera_id(),
+            frame.context().camera_buffer_timestamp(),
+            frame.context().detected().unwrap_or(false),
+            frame
+                .context()
+                .centroid()
+                .map(|(cx, cy)| PixelCenter::new(cx, cy)),
+        );
+
+        let _ = frame.clear_undistorted_image();
```
owens3364: `from` should not mutate its input data

its job is to provide a new `ContourOutput`

Skip the frame mutations and just return contour output as soon as you've created it

backend/src/pipeline/contour_output.rs
```diff
@@ -0,0 +1,174 @@
...
+        let _ = frame.clear_downsampled_image();
+
+        contour_output
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct MatchedContourPair {
+    left: ContourOutput,
```
owens3364: If we have left and right in the pair, do we need camera id in `ContourOutput`?

backend/src/pipeline/frame.rs
```diff
@@ -45,8 +45,8 @@ impl Frame {
         Self {
             raw_bytes_full_resolution: data,
             raw_full_resolution: resolution,
-            undistorted_image: OnceLock::new(),
-            downsampled_image: OnceLock::new(),
+            undistorted_image: RwLock::new(None),
```
owens3364, line 59: If it is possible to pre-allocate our `Mat`s here but *not* initialize the memory that they own, that would be fantastic.

If that looks painful, dw abt it, we can figure it out in another PR

backend/src/pipeline/frame.rs
```diff
@@ -59,29 +59,69 @@ impl Frame {
         self.raw_full_resolution
     }
 
-    pub fn undistorted_image(&self) -> Option<&Mat> {
-        self.undistorted_image.get()
+    pub fn undistorted_image(&self) -> Option<Mat> {
```
owens3364, line 73: Let's not do clones in reads - this will be a huge copy

owens3364: I believe `as_ref()` should work here no?

owens3364 (changes requested):

backend/src/pipeline/contour_output.rs
MaxWiesenfeld: The idea is that after computer_vision pipeline is done, the original frame can be sent somewhere else, but only a highly slimmed data structure needs to continue for yash's work. Since preparing the inputs to his optimize_trajectory requires savings many values, my thought was it is good to keep these cheaper structures.

backend/src/pipeline/contour_output.rs
MaxWiesenfeld: Just shedding no longer needed memory--that intermediate CV stage is not going to be used for anything else, no?

backend/src/pipeline/frame.rs
MaxWiesenfeld: Tbh don't understand the memory allocation very well and have been leaning on chat for help with it. Would say leave it for another PR if not critical. This one is also already bloated.

backend/src/pipeline/frame.rs
MaxWiesenfeld: Again, unfortunately not knowledgeable about memory management, but chat is making the case that handing out an as_ref of the lock is a direct read access which Rust rejects(?). Maybe take care of this in follow-up memory PR with above comment?

backend/src/pipeline/contour_output.rs
luke-venk: The caller can shed no longer needed memory on its own. `From` and `Into` are common STL Rust features for converting between types. Side effects are not expected for this.

backend/src/pipeline/frame.rs
luke-venk: Cloning a matrix is copying bitwise which is huge for this `Mat`. The `Frame` owns this memory already so borrowing is preferred/safe since it doesn't copy. If it doesn't compile then you can do this

luke-venk: Looks like using `as_ref()` for something that's behind a `RwLock` is a bit more difficult. Simply swapping `clone()` for `as_ref()` doesn't compile:

```
error[E0515]: cannot return value referencing function parameter `guard`
  --> backend/src/pipeline/frame.rs:77:31
   |
77 |             .and_then(|guard| guard.as_ref())
   |                               -----^^^^^^^^^
   |                               |
   |                               returns a value referencing data owned by the current function
   |                               `guard` is borrowed here
```

If we still want to use references, we'd have to refactor for the caller to have the `RwLockReadGuard` and deal with that, which is probably a pain. Or we could use `Arc`. Otherwise we could just leave it as clone.

luke-venk (approved):

PR #77, opened 2026-04-28
Centralize left/right camera assignment with CLI/manifest/repo-config resolver

RushilRandhar:
## Summary

Adds a single `CameraAssignment` resolver shared by live capture, the recorder, and replay so all three modes agree on which physical camera is FieldLeft vs FieldRight. Replaces the previous implicit-alphabetical-from-`get_camera_ids()` behavior with an explicit precedence chain that is overridable from the CLI and persistable per-machine via a repo-relative config file.

**Branched off `origin/test-camera-ingest`** because the replay/manifest infrastructure this work plugs into lives there, not on `main`. PR retargets to `main` so it lands once `test-camera-ingest` is merged (the diff shown will collapse appropriately).

## Precedence (first match wins)

1. `--left-camera-id` / `--right-camera-id` CLI flags. If only one is given, the other is inferred from the available pool (exactly two cameras).
2. Session manifest (`recording_session.json`) — replay only.
3. `camera_assignment.json` at the workspace root.
4. Panic with copy-paste-ready setup instructions (discovered camera IDs + sample JSON + path).

## What changed

- New [backend/src/camera/camera_assignment.rs](backend/src/camera/camera_assignment.rs) with the resolver + 8 unit tests.
- New `--left-camera-id` / `--right-camera-id` on `RealBackendArgs` and `CommonRecordArgs`.
- `main.prod.rs`, `main.dev.rs`, `record_from_both_cameras.rs`, and `replay_recorded_session.rs` all route through the resolver. Old hand-rolled alphabetical fallback in `replay_recorded_session::resolve_camera_mapping` deleted.
- `start_recorded_footage_pipelines` extended to forward CLI overrides into replay.
- `recording_session.json` now records the *resolved* assignment, not blind `camera_ids[0]/[1]`.
- New `//backend:dump_first_frames` binary writes the first frame from each camera as 8-bit grayscale PGM (no new crate deps; opens in macOS Preview) so physical orientation can be eyeballed.
- [camera_assignment.example.json](camera_assignment.example.json) at the workspace root documents the schema.

## Open questions / follow-ups

1. **`.gitignore` for `camera_assignment.json`?** Not committed in this PR. The committed `.example.json` is the schema reference; whether the actual config file should be gitignored (per-machine) or committed (team-wide) is a team call.
2. **macOS deployment-target mismatch on local builds.** On macOS 26 (Tahoe) with current homebrew OpenCV 4.13.0_7, the Bazel toolchain's hardcoded `-mmacosx-version-min=11.0.0` collides with OpenCV's `minos 26.0`, producing a linker error (`libopencv_dnn.dylib has version 26.0.0, which is newer than target minimum of 11.0`). This affects *every* binary that links OpenCV — including `record` from `main` — not anything in this PR. Worth a separate ticket to bump the toolchain's min-os to something like 14.0 or 15.0 so contributors on recent Macs don't have to fight homebrew. Confirmed by building only the rlib (`//backend:backend_lib_real_cameras`), which compiles cleanly.

## Test plan

- [ ] CI builds `//backend:prod_real_cameras`, `//backend:dev_real_cameras`, `//backend:record`, `//backend:dump_first_frames`, `//backend:tests`. *Local builds blocked by the OpenCV/min-os issue above; rlib confirmed clean locally.*
- [ ] CI runs `//backend:tests` (covers resolver precedence chain, replay manifest routing, replay CLI override, and the new `load_recorded_session` ordering test).
- [ ] **Manual on-rig verification (still needed)**: record a session with both cameras, confirm `recording_session.json` is written and contains the expected assignment, then run `//backend:dump_first_frames -- --footage-dir <session> --output-dir /tmp/check` and visually confirm `left_first.pgm` came from the camera physically mounted on the left side of the field.
- [ ] **Manual replay test (still needed)**: `//backend:prod_real_cameras -- --feed-footage-dir <session>` produces `pipeline: FieldLeft produced output frame ...` / `pipeline: FieldRight produced output frame ...`, and the assignment matches the manifest.
- [ ] **CLI override sanity check (still needed)**: same replay with `--left-camera-id <swapped>` deliberately flips left/right, no panics.


RushilRandhar:
Superseding with a 5-PR breakdown: (0) Bazel min-os toolchain fix, (1) SessionManifest, (2) replay infrastructure, (3) recorder PTP/barrier integration, (4) centralized assignment resolver. Each smaller and independently reviewable.

PR #79, opened 2026-04-28
Improve Recording Metrics Prints

### LINKED ISSUE #66: Make Frames Per Second Calculation More Robust (and clean up prints)
## Description
Right now, our camera frames per second metric which is printed at the end of acquisition isn't super robust. It doesn't count dropped frames and it probably should. Also I'm not sure that the correct elapsed time is used.

luke-venk:
Closes #66 

You made the point that our metrics should account for dropped frames. I made it so that our effective FPS print only considers the frames that actually made it to our camera (since that informs us more of our networking hardware woes) but also reported the dropped frames to provide more insights.

### DIFF
--- a/backend/src/camera/record/capture.rs
+++ b/backend/src/camera/record/capture.rs
@@ -66,7 +66,6 @@ pub fn send_action_command(
     group_key: u32,
     group_mask: u32,
 ) {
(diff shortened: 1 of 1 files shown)


RushilRandhar (approved):
Bet

PR #84, opened 2026-04-30
Add mog2 CV stage

MaxWiesenfeld:
## Justification
This PR adds the logic to convert mog2 as done in Python in Arushi's tune_cv.py file into a proper Rust pipeline stage. This is another step in the effort to complete the computer vision pipeline as set by #4 .

## Implementation
Copied logic from Python--filtering to only include mog2 morphology and related functions.

## Misc
Merge #74 first, then merge into this branch to reduce diff.


backend/src/computer_vision/mog2.rs
```diff
@@ -1,11 +1,129 @@
 use crate::pipeline::Frame;
+use opencv::core::{Mat, Point, Scalar, Size, BORDER_CONSTANT};
+use opencv::imgproc::{self, MORPH_CLOSE, MORPH_ELLIPSE, MORPH_OPEN, THRESH_BINARY};
+use opencv::prelude::*;
+use opencv::video;
+use std::cell::RefCell;
 
-/// The number of frames required for Mog2 to build the background model.
 pub const MOG2_HISTORY_FRAMES: usize = 300;
+const MOG2_VAR_THRESHOLD: f64 = 60.0;
```
luke-venk, line 9: can you put this to 80? that's the value in Owen's updated pipeline

backend/src/computer_vision/mog2.rs
```diff
@@ -1,11 +1,129 @@
...
+            MOG2_HISTORY_FRAMES as i32,
+            MOG2_VAR_THRESHOLD,
+            MOG2_DETECT_SHADOWS,
+        )
+        .expect("failed to create MOG2 background subtractor");
+
+        let kernel_open = imgproc::get_structuring_element(
+            MORPH_ELLIPSE,
+            Size::new(MORPH_OPEN_KERNEL_SIZE, MORPH_OPEN_KERNEL_SIZE),
+            Point::new(-1, -1),
```
luke-venk, line 34: Why do we need Point (-1, -1)? Don't recognize this code in Python

backend/src/computer_vision/mog2.rs
```diff
@@ -1,11 +1,129 @@
...
+        )
+        .expect("failed to create morphological open kernel");
+
+        let kernel_close = imgproc::get_structuring_element(
+            MORPH_ELLIPSE,
+            Size::new(MORPH_CLOSE_KERNEL_SIZE, MORPH_CLOSE_KERNEL_SIZE),
+            Point::new(-1, -1),
```
luke-venk, line 41: Same as https://github.com/luke-venk/mjolnir/pull/84/changes#r3171616090

backend/src/computer_vision/mog2.rs
```diff
@@ -1,11 +1,129 @@
...
+        Self {
+            bg_sub,
+            kernel_open,
+            kernel_close,
+        }
+    }
+
+    fn process_frame(&mut self, frame: Frame) -> Frame {
+        let Some(gray) = frame.downsampled_image() else {
+            return frame;
```
luke-venk, line 54: Note that we shouldn't fail silently here (by returning frame). In the future, check for the `use_4k` flag, and if it's false and downsampled image is none, panic.

backend/src/computer_vision/mog2.rs
```diff
@@ -1,11 +1,129 @@
...
+        // Background subtraction: learning_rate=-1 uses the MOG2 default schedule.
+        let mut fg_mask = Mat::default();
+        if self.bg_sub.apply(&gray, &mut fg_mask, -1.0).is_err() {
+            return frame;
+        }
 
-    frame
+        // Threshold to binary: eliminates shadow pixels (marked 127 by MOG2).
+        let mut fg_binary = Mat::default();
+        if imgproc::threshold(
```
luke-venk, line 65: Change to this structure, so it doesn't silently fail
```
if let Err(err) = imgprc::treshold(...){ 
    panic!("Thresholding for Mog2 failed: {}", err);
}
```

backend/src/computer_vision/mog2.rs
```diff
@@ -1,11 +1,129 @@
...
+            THRESH_BINARY,
+        )
+        .is_err()
+        {
+            return frame;
+        }
+
+        // Morphological open: removes small noise blobs.
+        let mut mask_opened = Mat::default();
+        if imgproc::morphology_ex(
```
luke-venk, line 79: Same as https://github.com/luke-venk/mjolnir/pull/84/changes#r3171667506

backend/src/computer_vision/mog2.rs
```diff
@@ -1,11 +1,129 @@
...
+            Scalar::default(),
+        )
+        .is_err()
+        {
+            return frame;
+        }
+
+        // Morphological close: fills holes inside the implement blob.
+        let mut mask_clean = Mat::default();
+        if imgproc::morphology_ex(
```
luke-venk, line 96: Same as https://github.com/luke-venk/mjolnir/pull/84/changes#r3171667506

backend/src/computer_vision/mog2.rs
```diff
@@ -1,11 +1,129 @@
...
+            return frame;
+        }
+
+        // Replace the downsampled image with the cleaned binary mask.
+        // The contour stage reads downsampled_image as its input mask.
+        frame
+            .clear_downsampled_image()
+            .expect("failed to clear downsampled image before writing mog2 output");
+        frame
+            .set_downsampled_image(mask_clean)
```
luke-venk, line 117: You can clear the downsampled image if you want, but we should store a separate `RwLock<Option<Mat>>` like `morphed_mog2_output` in `Frame`. The only reason it wasn't there before is because I wasn't adding stuff for stages we hadn't covered yet

luke-venk (changes requested):

PR #85, opened 2026-05-01
Input Into Trajectory Model Changes

yashj1579:
Hopefully everything should be good.


backend/src/math_triangulation/math_triangulation.rs
```diff
@@ -175,14 +171,13 @@ impl LeastSquaresProblem<f64, Dyn, Dyn> for TrajectoryProblem {
...
 //     dt: Time step (1/fps).
-//     g: Gravity vector.
-//     drag: Drag coefficient.
-//     pixel_sigma: Standard deviation of the pixel noise.
-//     physics_sigma: Standard deviation of the physics noise.
-//     omega_phys: Hyperparameter for the physics noise.
 // Returns: trajectory, position covariance, optimized drag, LM success flag.
-pub async fn optimize_trajectory(P_list: &[Matrix3x4<f64>], pixels: &[Vec<Vector2<f64>>], dt: f64, g: Option<Vector3<f64>>, drag: f64, pixel_sigma: f64, physics_sigma: f64, omega_phys: f64,) -> (Vec<Vector3<f64>>, DMatrix<f64>, f64, bool) {
-    let g = g.unwrap_or_else(|| Vector3::new(0.0, 0.0, -9.81));
+pub async fn optimize_trajectory(P_list: &[Matrix3x4<f64>], pixels: &[Vec<Vector2<f64>>], dt: f64) -> (Vec<Vector3<f64>>, DMatrix<f64>, f64, bool) {
```
luke-venk: It doesn't look `P_list` is hard-coded yet right? That's like one of the main things we need, could you please hard code that like how you did below?

backend/src/math_triangulation/math_triangulation.rs
```diff
@@ -175,14 +171,13 @@ impl LeastSquaresProblem<f64, Dyn, Dyn> for TrajectoryProblem {
 //     P_list: List of camera projection matrices. Should be gotten from camera calibration.
 //     pixels: List of pixel coordinates for each camera at each timestep. Should be gotten from CV.
 //     dt: Time step (1/fps).
-//     g: Gravity vector.
```
luke-venk, line 178: Could you please move all these comments you deleted to where you hard-coded the variables so people still know what they are?

backend/src/math_triangulation/mod.rs
```diff
@@ -10,4 +10,8 @@
 
 pub mod triangulation_output;
 
+#[path = "math_triangulation.rs"]
```
luke-venk: You could just keep the name `math_triangulation` I don't think that's confusing

luke-venk (changes requested):
Can you please revert the submodules to the versions on main?

backend/BUILD.bazel
```diff
@@ -38,6 +38,7 @@ rust_library(
         "@crates//:nalgebra",
         "@crates//:opencv",
         "@crates//:rand",
+        "@crates//:rand_distr",
```
luke-venk: I don't see this used anywhere, why is it here

luke-venk (changes requested):

backend/BUILD.bazel
yashj1579: we are adding gaussian noise to the test data in order to fabricate noisy data.

luke-venk: I ended up just adding it to the test dependencies instead of each library

backend/src/math_triangulation/mod.rs
luke-venk: Ahhh you didn't have to totally undo this file, just didn't have to name it something else. Turns out that not having this excludes all your code from the repo, so I had to re-add it

luke-venk (approved):
I ended up doing some changes myself. The big thing was that when you undid the changes to `mod.rs` it excluded all this from the repo which I had to fix. Other than that it works

PR #91, opened 2026-05-01
Include Camera PTP Offset in Global Time

### LINKED ISSUE #76: Find Difference Between Local PC Time and Camera PTP Time
Spin off ticket from [this comment](https://github.com/luke-venk/mjolnir/pull/55#discussion_r3142585458) on #55.

## Problem
Frames from cameras get not delivered instantaneously due to networking issues, like a delay of 2-5 seconds (now-ish, but perhaps a few seconds ago). When we receive a circle infraction, it's hard to say when in the throw trajectory that specifically happened.

We need a way to detect the difference between our local laptop time and the PTP assigned time on the cameras. If we figure this out, it'll be easier for us to know if an incoming circle infraction should be assigned to a specific throw or not.

## Work Needed
1. At startup, we should send a bunch of requests to camera ( search for `read_ptp_time_ns` for how to get PTP time from cameras in `backend/src/camera/record/capture.rs`). We can then do math to figure out the difference between local time and the camera's PTP time. Probably spam a few requests and take the average of these differences.
2. Create a struct in an appropriate location like `CircleInfractionTimestamps` that stores the 2 timestamps. Only bother creating this if the received signal was an infraction (i.e. ignore if not an infraction)
  a. The PC timestamp when the infraction arrives
  b. What the PTP timestamp *would* be (i.e. figure out what the time of infraction would be from the camera's perspective by adding whatever difference you found in step 1.)
3. Every time we get a circle infraction, create the structure in step 2. We now have all the time information we need from the perspective of the camera (comparing apples to apples), so we can later hard code some value to be like "if the infraction happened in the last 2 seconds, put it to the current infraction".

## Relevant files
- `backend/src/camera/record/capture.rs`
- `backend/src/circle_infractions_ingest/circle_infractions_thread.rs`: Refer to code beginning on line 84

owens3364:
Related to #76

### DIFF
--- a/backend/src/camera/aravis_utils.rs
+++ b/backend/src/camera/aravis_utils.rs
@@ -1,9 +1,10 @@
 use super::{BarrierResult, CancelableBarrier};
 use crate::camera::CameraIngestConfig;
+use crate::timing::global_time;
 use aravis::glib::translate::ToGlibPtr;
 use aravis::prelude::*;
 use aravis::{Aravis, Buffer, Camera, Stream};
-use aravis_sys::arv_camera_get_string;
+use aravis_sys::{arv_camera_get_string, arv_camera_get_integer};
 use glib::translate::*; // To convert high-level types to raw pointers
 use std::ffi::CString;
 use std::ptr;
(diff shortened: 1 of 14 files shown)


owens3364:
frontend/package.json patched to fix an obscure and rare bazel build error

backend/src/camera/aravis_utils.rs
```diff
@@ -60,6 +61,60 @@ fn unsafe_read_camera_boolean(camera: &Camera, node_name: &str) -> bool {
...
+        .expect("Failed to latch PTP dataset.");
+    unsafe_read_camera_integer(camera, "PtpDataSetLatchValue") as u64
+}
+
+/// This mutates the global time to include a PTP offset
+/// It makes 50 'what PTP time is it' calls to the camera
+/// Estimates the round trip time and offset
+/// Removes outliers
+/// And sets the PTP offset in global time to the average of the offsets after outliers are removed.
+/// 'Outlier' is classified as any sample with a rount-trip time >3x the minimum one
```
luke-venk: nit: "rount-trip" -> "round-trip"

backend/src/camera/aravis_utils.rs
```diff
@@ -60,6 +61,60 @@ fn unsafe_read_camera_boolean(camera: &Camera, node_name: &str) -> bool {
...
+/// Since it is hard to get good camera time if IO/OS/network jitter fucked our RTT
```
luke-venk, line 93: nit: could you specify that RTT is round trip time

frontend/package.json
```diff
@@ -25,5 +25,11 @@
     "eslint-config-next": "16.1.6",
     "tailwindcss": "^4",
     "typescript": "^5"
+  },
+  "pnpm": {
```
luke-venk, line 29: why would this be needed for this?

luke-venk (changes requested):
Super cool. Just some nits and I'll approve. Also once #90 is merged in the diff for this should decrease

frontend/package.json
owens3364: https://github.com/luke-venk/mjolnir/pull/91#issuecomment-4358928311

Those packages occsionally bring in binary deps that bazel fails to build, I don't totally understand why but I had a failing bazel build and this fixes it

We use this in trel3 as well

luke-venk: Thanks, sorry I didn't notice before that you mentioned it

luke-venk (approved):

PR #92, opened 2026-05-01
Circle infraction timestamping & real serving

### LINKED ISSUE #76: Find Difference Between Local PC Time and Camera PTP Time

owens3364:
Related to #76

### DIFF
--- a/backend/src/server/server.rs
+++ b/backend/src/server/server.rs
@@ -10,6 +10,7 @@ use axum::{
     response::IntoResponse,
     routing::{get, post},
 };
+use crate::timing::global_time;
 use crossbeam::channel::Receiver;
 use serde_json::json;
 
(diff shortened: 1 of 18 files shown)


backend/src/server/server.rs
```diff
@@ -54,16 +55,28 @@ async fn get_throw_type(State(state): State<AppState>) -> Json<GetThrowTypeRespo
...
-            Json(simulate_throw_event(throw_type))
+            simulate_throw_event(throw_type)
         }
+    };
+    let now_ns = global_time()
+        .camera_ptp_time_now_approximation_nanoseconds()
+        .unwrap_or_else(|| {
+            global_time().now_monotonic_in_nanoseconds_since_unix_epoch()
+        });
+    let ten_seconds_ago_ns = now_ns.saturating_sub(10_000_000_000);
```
luke-venk, line 73: Is the final plan to hard code the 10 second window or would we update it based on throw start/end using CV?

luke-venk (commented):
Not approving yet bc had some convos to resolve on #91, so approving now would just discard it once you change

backend/src/server/server.rs
owens3364: We would update it using throw start/end but we aren't there yet :/

luke-venk (approved):

PR #93, opened 2026-05-01
Implement Live Streaming into Pipelines

### LINKED ISSUE #86: Implement Live Stream To Feed Into Pipelines

owens3364:
Closes #86


backend/src/camera_ingest/live_dual_cam_ingest.rs
```diff
@@ -0,0 +1,338 @@
...
+use crate::computer_vision::mog2::MOG2_HISTORY_FRAMES;
+use crate::pipeline::{CameraId, Context, Frame};
+use aravis::{BufferStatus, CameraExt, StreamExt};
+use crossbeam::channel::{bounded, Receiver};
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+use std::thread;
+use std::time::{Duration, Instant};
+
+const MAX_QUEUED_FRAMES: usize = 10;
```
luke-venk: Is this the same number as `let capacity_per_channel: usize = 10;` in the [main files?](https://github.com/luke-venk/mjolnir/pull/93/changes#diff-dbb9de54b17409db01beb780c4faf54710506202e2b6a95431ba1e8b4166cedeR54)

If so, should we define as pub const in a shared file and just use that everywhere?

backend/src/camera_ingest/live_dual_cam_ingest.rs
```diff
@@ -0,0 +1,338 @@
...
+
+pub fn begin_live_dual_cam_ingest(
+    left_cam_id: String,
+    right_cam_id: String,
+    exposure_time_us: f64,
+) -> (Receiver<Frame>, Receiver<Frame>) {
+    let _ = initialize_aravis();
+
+    let left_cam_ingest_config = CameraIngestConfig {
```
luke-venk, line 22: Wouldn't it be more robust to use a function like `from_record_one_args` [defined here](https://github.com/luke-venk/mjolnir/blob/main/backend/src/camera/config.rs#L27)?

You could make the argument that these numbers should be hard coded for dual cam ingest, but you could still hard code in a function like that in `CameraIngestConfig` to make it more organized. Something like `fn for_dual_cam_ingest()` or whatever you think is clean

backend/src/camera_ingest/live_dual_cam_ingest.rs
```diff
@@ -0,0 +1,338 @@
...
+        camera_id: left_cam_id,
+        exposure_time_us,
+        frame_rate_hz: 30.0,
+        resolution: AtlasATP124SResolution::Full,
+        num_buffers: 16,
+        timeout_ms: 5000,
+        restart_requested: false,
+    };
+
+    let right_cam_ingest_config = CameraIngestConfig {
```
luke-venk, line 32: Same as https://github.com/luke-venk/mjolnir/pull/93/changes#r3173972505

backend/src/camera_ingest/live_dual_cam_ingest.rs
```diff
@@ -0,0 +1,338 @@
+use crate::camera::aravis_utils::{
```
luke-venk, line 1: Should we go ahead and just delete the original `backend/src/camera_ingest/camera_ingest.rs` file that Gautam was supposed to make, since Rushil handled feeding recorded footage and you handled live stream?

luke-venk (commented):
Might be better to test on the cameras (in Alex's lab) before merging

backend/src/camera_ingest/live_dual_cam_ingest.rs
owens3364: Mmmm in this case we don't need validation because nearly everything is hard-coded (and should be). Only things not hardcoded are camera ids/names, which we can't really validate until we try to connect to them, and the exposure time us, which I will add validation on before this function is called. But yeah, since so few CLI args and so much hardcoded, IMO extra validation funcs unnecessary here.

backend/src/camera_ingest/live_dual_cam_ingest.rs
owens3364: https://github.com/luke-venk/mjolnir/pull/93#discussion_r3174895206

backend/src/camera_ingest/live_dual_cam_ingest.rs
owens3364: Actually we hardcode exposure time too

There are no cli args for this so CLI arg parsing seems unnecessary IMO

backend/src/camera_ingest/live_dual_cam_ingest.rs
luke-venk: Shouldn't the lines using this just refer to [CAPACITY_PER_CROSSBEAM_CHANNEL](https://github.com/luke-venk/mjolnir/pull/93/changes#diff-3d3d7ecb11a70a50a2d3cb45a2ddad997f86a8fba5c6879abfc853ce5562dc1eR14) and this be deleted? Or are they not the same thing?

backend/src/camera_ingest/live_dual_cam_ingest.rs
luke-venk: Yeah that makes a lot of sense

luke-venk (approved):

